PR #5, opened 2022-10-23
close #1 user registration with social

### LINKED ISSUE #1: Authentication with social

kimleang123:


channainfo:
@kimleang123 let's **namespace** the class in our gem.
UserRegistrationWithSocial -> Commissioner:: UserRegistrationWithSocial

Every models/services/controllers should lay under **commissioner** directory.

app/controllers/spree/api/v2/commissioner/user_registration_with_socials_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Spree::Api::V2::Commissioner::UserRegistrationWithSocialsController
+    user_context = ::UserRegistrationWithSocial.call(id_token: params[:id_token])
```
channainfo, line 2: @kimleang123 why no action method here?

db/migrate/20221021043620_create_user_identity_providers.rb
```diff
@@ -0,0 +1,16 @@
+class CreateUserIdentityProviders < ActiveRecord::Migration[6.1]
+  def change
+    create_table :user_identity_providers do |t|
```
channainfo, line 3: the table name should be **commissioner_user_identity_providers**. Prefixing the every table with the module name **commissioner_**

lib/spree_cm_commissioner/factories.rb
```diff
@@ -4,3 +4,26 @@
   # Example adding this to your spec_helper will load these Factories for use:
   # require 'spree_cm_commissioner/factories'
 end
+
+FactoryBot.define do
```
channainfo, line 8: split factory definition to its own file

channainfo (changes requested):

PR #16, opened 2022-11-05
Close #11 vendor logo

### LINKED ISSUE #11: Vendor Logo
we may need vendor logo in case we need to display on partnership page on bookmebus or vtenh. There are many cases that we still need it. 

Requirement: 
- only 1 image (default 250x250px)
- spree shall generate variants for different size of image

kimsrung:
image


app/models/spree/vendor_logo.rb
```diff
@@ -0,0 +1,18 @@
+module Spree
+  class VendorLogo < Asset
+    include Spree::Image::Configuration::ActiveStorage
+    include Rails.application.routes.url_helpers
+
+    def styles
+      self.class.styles.map do |_, size|
```
channainfo, line 7: we don't need all the styles. we can just use two styles: thumb( 60x60 ) and small ( 180x180)

channainfo (changes requested):

PR #18, opened 2022-11-12
12 vendor min max price

kimsrung:
- add field min_price and max_price to vendor
- add cronjob to update vendor min & max price every 24 hours
- add spec
- refactor custom class or module with namespace SpreeCmCommissioner


channainfo (changes requested):

Gemfile
```diff
@@ -13,4 +13,14 @@ gem 'spree_multi_vendor'
 
 gem 'rails-controller-testing'
 
+# for cron jobs
+gem 'whenever', require: false
```
channainfo, line 17: let's remove this. We will not use a cron but a scheduler instead ( sidekiq scheduler, sidekiq cron, or perhaps an AWS ECS scheduler to run a rake task ). 

The most scalable way is to have it as a **rake task** and then launch an AWS ECS scheduler). If it is a light-weighted task we can just use the Sidekiq scheduler to run. Cron requires a permanent instance and shared resource and becomes an indeterministic problem.

let's create a rake task and add a section in README under Scheduler sections. 
For each scheduler entry to describe:
- The job name
- Description of the Job
-  rake task command 
- How frequently do we want to run it

spec/spec_helper.rb
```diff
@@ -5,7 +5,20 @@
...
 require 'spree_multi_vendor/factories'
+require 'spree_travel_core/factories'
+
+require "interactor"
 
 # Requires supporting ruby files with custom matchers and macros, etc,
 # in spec/support/ and its subdirectories.
 Dir[File.join(File.dirname(__FILE__), 'support/**/*.rb')].sort.each { |f| require f }
+
+Shoulda::Matchers.configure do |config|
```
channainfo, line 16: let's move this block to its own file shoulda.rb and put it under spec/support to make it auto loaded.

channainfo (changes requested):

channainfo (approved):

Issue #19, opened 2022-11-21
Product Properties


theachoem:
# Product Property

## Breakdown
- [x] #57 
  - [x] Add `is_master` field to option type. Once set, can't be updated.
  - [x] Added `master` as name prefix for master option types.
- [x] #59 
  - [x] Exclude master option types from variant form.
  - [x] Add page to toggle option values to master variant.
- [x] #60
- [ ] Add master option type APIs
- [ ] Change option value input UI base on attr_type
## UI objective
| |
| - |
| ![](https://user-images.githubusercontent.com/29684683/202959752-fc243b3a-3c83-4ec7-b938-5da725a23d86.png) |

theachoem:
To follow spree structure, we will add Spree::OptionValueProduct to connect between option value and product.
|  |
| - |
| ![](https://user-images.githubusercontent.com/29684683/203465093-c6f94913-c1d6-4360-a283-7a0c86ca2038.png) |

PR #20, opened 2022-11-21
Search and return inventory

kimsrung:
Thing to do:
1. seed data for provinces: rake data:seed_kh_provinces
2. seed data for option value: rake data:seed_kh_location_option_values

API: {{apiServerUrl}}/api/v2/storefront/search?province_id={province_id}

### DIFF
--- a/app/controllers/spree/api/v2/storefront/search_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/search_controller.rb
@@ -1,0 +1,32 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class SearchController < ::Spree::Api::V2::ResourceController
+
+          private
+
+          def collection
+            @collection ||= province_id.present? ? scope.ransack(option_values_presentation_eq: province_id).result(distinct: true) : scope
+          end
+
+          def province_id
+            params[:province_id]
+          end
+
+          def scope
+            model_class.active
+          end
+
+          def model_class
+            ::Spree::Vendor
+          end
+
+          def collection_serializer
+            Spree::V2::Storefront::VendorSerializer
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/models/spree/product_decorator.rb
+++ b/app/models/spree/product_decorator.rb
@@ -1,9 +1,28 @@
 module Spree
   module ProductDecorator
     def self.prepended(base)
+      base.has_many :option_values, through: :option_types
       base.has_many :prices_including_master, -> { order('spree_variants.position, spree_variants.id, currency') }, source: :prices, through: :variants_including_master
       base.scope  :min_price, -> (vendor) { joins(:prices_including_master).where(vendor_id: vendor.id).minimum('spree_prices.price').to_f }
       base.scope  :max_price, -> (vendor) { joins(:prices_including_master).where(vendor_id: vendor.id).maximum('spree_prices.price').to_f }
+      base.after_commit :create_location
+    end
+
+    private
+    def create_location
+      location = vendor.stock_locations.first
+      raise ("Missig state in stock location for vendorId: #{vendor.id}") if location&.state_id.nil?
+
+      option_type = Spree::OptionType.find_or_create_by(name: 'location', presentation: 'Location', attr_type: 'state_selection')
+      option_value = option_type.option_values.find_or_create_by(presentation: location.state_id, name: location.name)
+
+      # append option type location to product
+      self.option_types << option_type if option_type_ids.exclude?(option_type.id)
+
+      # create option_value location to variants
+      variants.each { |v|
+        v.option_value_variants.find_or_create_by(option_value_id: option_value.id)
+      }
     end
   end
 end
--- a/lib/tasks/kh_location_option_values.rake
+++ b/lib/tasks/kh_location_option_values.rake
@@ -1,0 +1,14 @@
+namespace :data do
+  desc 'Seeds location option values'
+  task seed_kh_location_option_values: :environment do
+    location_type = Spree::OptionType.find_or_initialize_by(name: 'location', presentation: 'Location', attr_type: 'state_selection')
+    cambodia = Spree::Country.find_or_create_by(abbr: 'Cambodia')  # run `rake data:seed_kh_provinces` first
+
+    cambodia.states.each do |province|
+      location_type.option_values.find_or_initialize_by(name: province.name, presentation: province.id)
+    end
+    location_type.save
+
+    p "Created #{location_type.option_values.count} location option values"
+  end
+end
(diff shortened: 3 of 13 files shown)


channainfo:
@kimsrung add rake commands to README under the Seed sub section.

channainfo (changes requested):

app/models/spree/product_decorator.rb
```diff
@@ -1,9 +1,28 @@
...
+      base.after_commit :create_location
+    end
+
+    private
+    def create_location
+      location = vendor.stock_locations.first
+      raise ("Missig state in stock location for vendorId: #{vendor.id}") if location&.state_id.nil?
+
+      option_type = Spree::OptionType.find_by!(name: 'location', presentation: 'Location', attr_type: 'state_selection')
+      option_value = option_type.option_values.find_by!(state_id: location.state_id)
```
channainfo, line 17: we don't need to have state_id in the option_values. This field in which is very little in use and other records will carry on with. we can just store the state_id in the **value** field

db/migrate/20221121012928_add_state_id_to_option_value.rb
```diff
@@ -0,0 +1,5 @@
+class AddStateIdToOptionValue < ActiveRecord::Migration[6.1]
+  def change
+    add_column :spree_option_values, :state_id, :bigint
```
channainfo, line 3: Let's remove this.

channainfo (changes requested):

lib/tasks/kh_provinces.rake
```diff
@@ -0,0 +1,38 @@
+#rake db:seed:file_name
+
+namespace :data do
+  desc "Seeds provinces"
+  task seed_kh_provinces: :environment do
+    cambodia = Spree::Country.find_by(name: 'Cambodia')
+
+    cambodia.states.create!([
```
channainfo, line 8: @kimsrung let's also make this idempotent. It is ok to do them one by one without bulk insert since it is relatively small. We want to make sure they are safe to run multiple times.

spec/models/spree/product_spec.rb
```diff
@@ -48,4 +51,47 @@
       end
     end
   end
+
+  describe 'callback #create_location' do
```
channainfo, line 55: let's just use solely #create_location. You can either remove the **callback** text or create a nested block:

```
description 'callback' do
   description '#create_location' do

```

lib/tasks/kh_location_option_values.rake
```diff
@@ -0,0 +1,16 @@
...
+
+namespace :data do
+  desc 'Seeds location option values'
+  task seed_kh_location_option_values: :environment do
+    location_type = Spree::OptionType.find_or_create_by!(name: 'location', presentation: 'Location', attr_type: 'state_selection')
+    cambodia = Spree::Country.find_by(name: 'Cambodia')
+    provinces = cambodia.states
+
+    provinces.each do |province|
+      location_type.option_values.find_or_create_by!(name: province.name, presentation: province.id)
```
channainfo, line 11: let's use find_or_initalize_by to make it idempotent.

lib/tasks/kh_location_option_values.rake
```diff
@@ -0,0 +1,16 @@
+#rake db:seed:file_name
+
+namespace :data do
+  desc 'Seeds location option values'
+  task seed_kh_location_option_values: :environment do
+    location_type = Spree::OptionType.find_or_create_by!(name: 'location', presentation: 'Location', attr_type: 'state_selection')
```
channainfo, line 6: let's use find_or_initalize_by to make it idempotent.

app/controllers/spree/api/v2/storefront/search_controller.rb
```diff
@@ -0,0 +1,33 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class SearchController < ::Spree::Api::V2::ResourceController
+
+          private
+
+          def collection
+            @collection ||= scope.ransack(stock_locations_state_id_eq: province_id).result(distinct: true)
```
channainfo, line 10: let's make province id optional. if the province is not provided we return all.

app/controllers/spree/api/v2/storefront/search_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+          end
+
+          def province_id
+            raise ActionController::ParameterMissing if params[:province_id].blank?
```
channainfo, line 14: for now, we can just ignore this line.

channainfo (changes requested):

theachoem (approved):

theachoem (approved):
Just check whether it not conflict with our discussed flow here: https://github.com/channainfo/commissioner/issues/19

app/models/spree/product_decorator.rb
```diff
@@ -1,9 +1,28 @@
...
       base.scope  :max_price, -> (vendor) { joins(:prices_including_master).where(vendor_id: vendor.id).maximum('spree_prices.price').to_f }
+      base.after_commit :create_location
+    end
+
+    private
+    def create_location
+      location = vendor.stock_locations.first
+      raise ("Missig state in stock location for vendorId: #{vendor.id}") if location&.state_id.nil?
+
+      option_type = Spree::OptionType.find_by!(name: 'location', presentation: 'Location', attr_type: 'state_selection')
```
channainfo, line 16: always use ** find_or_initialize_by** instead -- create if it does not exist, otherwise, read. 

This will make sure we will be able to run the code without waiting for the rake task to have been run ahead, especially when new data is added and the rake task has already run before.

app/models/spree/product_decorator.rb
```diff
@@ -1,9 +1,28 @@
...
+      option_value = option_type.option_values.find_by!(presentation: location.state_id)
```
channainfo, line 17: use find_by_or_initialize instead. Create if it does not exist, otherwise, read.

lib/tasks/kh_location_option_values.rake
```diff
@@ -0,0 +1,14 @@
+namespace :data do
+  desc 'Seeds location option values'
+  task seed_kh_location_option_values: :environment do
+    location_type = Spree::OptionType.find_or_initialize_by(name: 'location', presentation: 'Location', attr_type: 'state_selection')
+    cambodia = Spree::Country.find_by!(name: 'Cambodia')  # run `rake data:seed_kh_provinces` first
```
channainfo, line 5: This case is a bit more specific but let's just use find_or_initialize_by here too following the same principle above. Let's use the **abbr** field instead. The **name** might be subjected to translation or use Country.cambodia

channainfo (changes requested):

lib/tasks/kh_location_option_values.rake
kimsrung: Spree doesn't have **abbr** field, it uses as **iso**. I will update that. But I just want to prefer to use ** find_or_create_by** instead

PR #71, opened 2022-12-23
Close #70 search product in stock

### LINKED ISSUE #70: Search product

kimsrung:


channainfo:
@kimsrung when searching for the products in stock we need the date range. The logic would be The total permanent stock available - total booking.

By default, one variant has one permanent stock.

kimsrung:
yes, bong. I think it's an old one. I actually updated in local already. Will push again.

app/services/spree_cm_commissioner/vendor_search.rb
```diff
@@ -1,17 +1,44 @@
...
       prepare(params)
     end
 
     def call
       @vendors = base_elasticsearch
     end
 
+    def variants_with_available_stock
+      # all_variants
+      option_type= Spree::OptionType.find_by name: 'location'
```
channainfo, line 18: Great solution. We can use this algorithm to optimize for scalability to offload the calculation to the DB engine and cache. 

You can try to migrate this to Postgresql CTE and then convert Active record query.

db/migrate/20221230053022_add_column_from_date_and_to_date_to_spree_orders.rb
```diff
@@ -0,0 +1,6 @@
+class AddColumnFromDateAndToDateToSpreeOrders < ActiveRecord::Migration[6.1]
+  def change
+    add_column :spree_orders, :from_date, :datetime
+    add_column :spree_orders, :to_date, :datetime
+  end
```
channainfo, line 5: double-check if an order has any existing field for the purchase date that we can use and add index multi fields index to speed up the query

channainfo (changes requested):

db/migrate/20221230053022_add_column_from_date_and_to_date_to_spree_orders.rb
kimsrung: in order, we have only fields: completed_at, approved_at, canceled_at... So, I will add index on from_date and to_date in order to improve query

app/models/spree/order_decorator.rb
```diff
@@ -0,0 +1,14 @@
+module Spree
+  module OrderDecorator
+    def self.prepended(base)
+
+      base.scope :date_range, ->(from_date, to_date) do
+        (from_date.blank? || to_date.blank?) ? all :
+          where("(from_date <= ? AND to_date > ?) OR (from_date <= ? AND to_date > ?)", from_date, from_date, to_date, to_date)
```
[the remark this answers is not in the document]
kimsrung: I change condition to this bong:
`base.scope :date_range, ->(from_date, to_date) { where(["from_date <= ? AND to_date > ?", to_date, from_date])}`

it's correct

PR #78, opened 2022-12-28
close #76 add product type

### LINKED ISSUE #76: Add product type type
image

theachoem:
# Product Type
Since we still not clear about what are [product types] to put in **enum**, using a table may be much easier but require more resources.

In this PR, we add new table called cm_product_types instead of enum with ERD below:

|  | 
| - |
| image |


theachoem:
Close due to we have to use enum instead

PR #80, opened 2022-12-28
close #73 add fields to spree user

### LINKED ISSUE #73: Fetch my account API

leak17:
pic4
pic3
pic2
pic1


app/controllers/spree/admin/users_controller_decorator.rb
```diff
@@ -0,0 +1,27 @@
+module Spree
+  module Admin
+    module UsersControllerDecorator
+      def self.prepended(base)
```
kimsrung, line 4: if it's empty, we don't need to add that

app/controllers/spree/admin/users_controller_decorator.rb
```diff
@@ -0,0 +1,27 @@
...
+          :password,
+          :password_confirmation,
+          :first_name,
+          :last_name,
+          :dob,
+          :gender,
+          :use_billing, 
+          spree_role_ids: [],
+          ship_address_attributes: permitted_address_attributes,
+          bill_address_attributes: permitted_address_attributes
```
kimsrung, line 19: I think if we wrote like this, it could lead to missing attributes from other gem. You can check gem spree_multi_vendor, it's also adding some new attributes to user_params as well. Just double check, if we are not missing out or override other attributes

image

app/controllers/spree/admin/users_controller_decorator.rb
```diff
@@ -0,0 +1,21 @@
...
+                                  :email,
+                                  :password,
+                                  :password_confirmation,
+                                  :first_name,
+                                  :last_name,
+                                  :dob,
+                                  :gender,
+                                  :use_billing,
+                                  spree_role_ids: [],
+                                  vendor_ids: [],
```
kimsrung, line 15: you don't need to add vendor_ids, ship_address_attributes.... etc..... just add your new fields here

leak17: we also need to override on spree_multi_vendor

kimsrung:
@leak17 : I think we can remove users_controller_decorator file. Since we only need to change attribute params, you can do in spree_permitted_attributes_decorator

image

app/controllers/spree/admin/users_controller_decorator.rb
```diff
@@ -0,0 +1,12 @@
...
+    super | [
+      :first_name,
+      :last_name,
+      :dob,
+      :gender
+    ]
+  end
+end
+
+Spree::Admin::UsersController.prepend Spree::Admin::UsersControllerDecorator
```
kimsrung, line 12: you can delete this file

PR #82, opened 2022-12-30
close #76 add product type with enum

### LINKED ISSUE #76: Add product type type

theachoem:
## Migrations
- [x] Added primary_product_type to vendor
- [x] Added product_type to product
- [x] Added product_type to prototype

## Logic changes
- Min/max vendor price will be updated based on `primary_product_type`.
  For example, if vendor.primary_product_type is `property`, then only variants with product_type `property` will trigger the update.

## Views
- [x] Added **product_types** collection_select to vendor view
- [x] Added **product_type** collection_select to prototype
- [x] Added **product_types** collection_select product view 
  - On `/edit`, This field is read only
  - On `/new`, product type will be has initial value base on `prototype.product_type`

image

### DIFF
--- a/app/models/concerns/spree_cm_commissioner/product_type.rb
+++ b/app/models/concerns/spree_cm_commissioner/product_type.rb
@@ -1,0 +1,15 @@
+# :spree_vendors
+# :spree_products, :spree_prototypes
+
+module SpreeCmCommissioner
+  module ProductType
+    extend ActiveSupport::Concern
+
+    PRODUCT_TYPES = %i[accommodation service ecommerce]
+
+    included do
+      enum product_type: PRODUCT_TYPES if column_names.include?("product_type")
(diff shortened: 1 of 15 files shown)


app/models/concerns/spree_cm_commissioner/product_type.rb
```diff
@@ -0,0 +1,15 @@
+# :spree_vendors
+# :spree_products, :spree_prototypes
+
+module SpreeCmCommissioner
+  module ProductType
+    extend ActiveSupport::Concern
+
+    PRODUCT_TYPES = %i[property service]
```
channainfo, line 8: - let us rename "property" to "accommodation". "property" is confusing when communicating especially the name has special meaning in programming and existing data design.
- let's add another one for "e-commerce" as well.

channainfo (changes requested):

PR #86, opened 2023-01-02
close #85 improve attr_type form for option value

### LINKED ISSUE #85: Enhance [attr_type] form for option value
References:
- https://www.npmjs.com/package/react-ui-form-builder-attachment

theachoem:
| | |
| - | - |
| Integer |  image |
| | |
| Float | image |
| | |
| Boolean | image |

### DIFF
--- a/app/models/concerns/spree_cm_commissioner/attr_type_validation.rb
+++ b/app/models/concerns/spree_cm_commissioner/attr_type_validation.rb
@@ -1,0 +1,31 @@
+module SpreeCmCommissioner
+  module AttrTypeValidation
+    extend ActiveSupport::Concern
+    
+    included do
+      def self.act_as_presentation
+        validates :presentation, numericality: { only_integer: true }, if: :integer
+        validates :presentation, numericality: { only_float: true }, if: :float
+        validates :presentation, inclusion: %w[1 0], if: :boolean
+      end
+    end
+
+    def option_type
+      @option_type ||= Spree::OptionType.find_by(id: option_type_id)
+    end
+
+    private
+
+    def integer
+      option_type&.attr_type == "integer"
+    end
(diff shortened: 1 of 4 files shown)


app/models/concerns/spree_cm_commissioner/attr_type_validation.rb
```diff
@@ -0,0 +1,31 @@
...
+    included do
+      def self.act_as_presentation
+        validates :presentation, numericality: { only_integer: true }, if: :integer
+        validates :presentation, numericality: { only_float: true }, if: :float
+        validates :presentation, inclusion: %w[1 0], if: :boolean
+      end
+    end
+
+    def option_type
+      @option_type ||= Spree::OptionType.find_by(id: option_type_id)
```
kimsrung, line 14: is it necessary to specify find_by id? because method find would take id automatically Spree::OptionType.find(option_type_id).

app/models/concerns/spree_cm_commissioner/attr_type_validation.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    private
+
+    def integer
+      option_type.attr_type == "integer"
```
kimsrung, line 20: just in case option_type is nil then it would cause an error here. we could use this option_type&.attr_type to prevent error

app/models/concerns/spree_cm_commissioner/attr_type_validation.rb
theachoem: - `find` will raise error when can't find it. Anyways, it's also not really necessary to put `find_by` since option value always has option type.

- But for safely, we use `find_by` will give nil instead of raise

PR #164, opened 2023-02-06
close #152 add promoted option values to vendor

chheng05:


app/models/spree/vendor_decorator.rb
```diff
@@ -18,12 +18,16 @@ def self.prepended(base)
...
 
       base.has_many :vendor_kind_option_values,
-        through: :option_value_vendors, source: :option_value
+        through: :option_value_vendors, source: :option_value 
 
       base.has_one  :logo, as: :viewable, dependent: :destroy, class_name: 'SpreeCmCommissioner::VendorLogo'
 
       base.after_save :update_state_total_inventory, if: :saved_change_to_total_inventory?
 
+      def promoted_option_value_ids
```
kimsrung, line 27: we can create a relationship for promoted_option_values, in case we may use it

```
base.has_many :promoted_option_values,-> { joins(:option_type).where('option_type.promoted' =>  true) },
        through: :option_value_vendors, source: :option_value
```
and for method  promoted_option_value_ids, we can pluck id from that

```
def promoted_option_value_ids
        promoted_option_values.pluck(:id)
      end
```

kimsrung:
please resolve conflict and update commit message to close #issue_id

theachoem:
@channainfo @kimsrung Done bong!

PR #174, opened 2023-02-13
close #173 fix undefine-method-each-file-for-assets

### LINKED ISSUE #173: Rails.application.assets.each_file : NoMethodError (undefined method `each_file' for nil:NilClass)
In production, if gem user doesn't set compile to true, it would return error.

```s
NoMethodError (undefined method `each_file' for nil:NilClass):
```

theachoem:


theachoem:
Demo: 
https://central-market-server.herokuapp.com/admin/vectors/icons

```
admin_dev@cm.com
superDEV@1
```

PR #179, opened 2023-02-14
close #178 require delivery only on product type ecommerce digital

### LINKED ISSUE #178: Require delivery only on product type ecommerce & digital

theachoem:
To get order.required_delivery?, we select only ecommece items and each of them is not digital. So, other product type such as accommodation & service will required delivery by default.

### DIFF
--- a/app/models/spree_cm_commissioner/order_decorator.rb
+++ b/app/models/spree_cm_commissioner/order_decorator.rb
@@ -1,0 +1,17 @@
+module SpreeCmCommissioner
+  module OrderDecorator
+    # required only in one case, 
+    # some of line_items are ecommerce & not digital.
+    def delivery_required?
+      contain_non_digital_ecommerce?
+    end
+
+    def contain_non_digital_ecommerce?
+      line_items.select {|item| item.ecommerce? && !item.digital? }.size > 0
+    end
+  end
+end
+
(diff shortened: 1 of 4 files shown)


app/models/spree_cm_commissioner/order_decorator.rb
```diff
@@ -0,0 +1,16 @@
+module SpreeCmCommissioner
+  module OrderDecorator
+    def delivery_required?
+      !ecommerce_digital?
+    end
+
+    def ecommerce_digital?
+      ecommerce_line_items = line_items.select {|item| item.ecommerce? } 
+      ecommerce_line_items.all?(&:digital?)
```
channainfo, line 9: it can be reduced to this:
```
line_items.select {|item| item.ecommerce? && !item.digital? }.size > 0
```

app/models/spree_cm_commissioner/order_decorator.rb
```diff
@@ -0,0 +1,16 @@
+module SpreeCmCommissioner
+  module OrderDecorator
+    def delivery_required?
+      !ecommerce_digital?
```
channainfo, line 4: I think it is more obvious to call it: **contain_non_digital_ecommerce?**

```
# overriden
def delivery_required?
  contain_non_digital_ecommerce?
```

channainfo (changes requested):

channainfo (approved):

PR #186, opened 2023-02-16
156 enhance rubocop

kimsrung:
image


.rubocop-performance.yml
```diff
@@ -0,0 +1,53 @@
+###########################################################
+#################### RuboCop Performance ##################
+###########################################################
+
+# You can find all configuration options for rubocop-performance here: https://docs.rubocop.org/rubocop-performance/
+
+Performance/AncestorsInclude: # (new in 1.7)
+  Enabled: true
```
channainfo, line 8: let's keep the config file minimal. Let's remove those cops with Enabled: true, we don't need them ( keep the default cop behavior).
Most likely we only need to negate the with Enabled: false. we will slowly adapt more cops along the way.

channainfo (changes requested):

channainfo:
Run rubocop locally and fix failed cop accordingly.

channainfo (approved):

PR #188, opened 2023-02-16
close #187 add taxon icon

### LINKED ISSUE #187: Add taxon icon

theachoem:
# Tasks
- [x] Create SpreeCmCommissiner:Asset as base class for all asset
- [x] Create SpreeCmCommissiner:TaxonCategoryIcon
- [x] Add shared view: asset_field
- [x] Add icon field view by using asset_field
- [x] Replace old head banner view asset_field

image

### DIFF
--- a/app/serializers/spree_cm_commissioner/v2/storefront/category_icon_serializer.rb
+++ b/app/serializers/spree_cm_commissioner/v2/storefront/category_icon_serializer.rb
@@ -1,0 +1,8 @@
+module SpreeCmCommissioner
+  module V2
+    module Storefront
+      class CategoryIconSerializer < AssetSerializer
+      end
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


app/serializers/spree_cm_commissioner/v2/storefront/category_icon_serializer.rb
```diff
@@ -0,0 +1,8 @@
+module SpreeCmCommissioner
+  module V2
+    module Storefront
+      class CategoryIconSerializer < AssetSerializer
```
kimsrung, line 4: don't we need to set_type for serializer?

theachoem: It use class type bong

theachoem: It will be type: category_icon

theachoem: I'm thinking of either put & not put them. But i think it will be more consistent this way as it unique base on class name.

kimsrung: can you check, in case we want to use specific name, can we put here and it's still working?

PR #204, opened 2023-02-27
close #202 validate merchant users

### LINKED ISSUE #202: Validate merchant users
Ensure the user belongs to the merchant by adding another constraint in the before_action: required_merchant_user!

Tykeaboyloy:
image
image


channainfo:
@Tykeaboyloy let's use default_url_options for the vendor_id and if there is no vendor_id in the params, get the first vendor as the current vendor.

Tykeaboyloy:
> @Tykeaboyloy let's use default_url_options for the vendor_id and if there is no vendor_id in the params, get the first vendor as the current vendor.

sorry bong, what does default_url_options refer to?

channainfo:
> 

@Tykeaboyloy plz take a look at this doc: https://guides.rubyonrails.org/action_controller_overview.html#default-url-options. You might need to do the same in your merchant controller using vendor_id.

Issue #182, opened 2023-02-15
Add vendor nearby location 


panhachom:
- [x] #207 
- [x] #211 
- [x] #213

PR #207, opened 2023-02-28
Close #206 get nearby places from google api

### LINKED ISSUE #206: get nearby places from google api

panhachom:
### Task Breakdown 
- Get nearby places from google API (min : 20 places , max : 60 places)
- Calculate  distance between places and vendor 
### Demo
[Screencast from 03-06-2023 09:34:53 AM.webm](https://user-images.githubusercontent.com/92453740/223007086-f1be89e0-6f5f-4e23-a5c7-1635a4b00a1b.webm)

### DIFF
--- a/app/controllers/spree/admin/nearby_places_controller.rb
+++ b/app/controllers/spree/admin/nearby_places_controller.rb
@@ -8,13 +8,8 @@ module Spree
       end
 
       def new
-        SpreeCmCommissioner::NearbyPlaceCreator.call(params: permitted_resource_params, vendor: @vendor)
-      end
-
-      private
-
-      def load_vendor
-        @vendor ||= Spree::Vendor.find_by(slug: params[:vendor_id])
+        context = SpreeCmCommissioner::VendorNearbyPlaceBuilder.call(vendor: @vendor)
+        @nearby_places = context.nearby_places
       end
 
       def model_class
@@ -25,20 +20,18 @@ module Spree
         'spree_cm_commissioner_nearby_places'
       end
 
-      def resource_serializer
-        Spree::V2::Storefront::NearbyPlaceSerializer
-      end
-
-      def collection_serializer
-        Spree::V2::Storefront::NearbyPlaceSerializer
-      end
-
       def collection_url(options = {})
         admin_vendor_nearby_places_url(options)
       end
 
       def permitted_resource_params
         params.permit(nearby_places: %i[name reference lat lon vicinity icon url rating formatted_address address_components])
+      end
+
+      private
+
+      def load_vendor
+        @vendor ||= Spree::Vendor.find_by(slug: params[:vendor_id])
       end
     end
   end
--- a/app/controllers/spree/admin/vendors_controller_decorator.rb
+++ b/app/controllers/spree/admin/vendors_controller_decorator.rb
@@ -3,9 +3,19 @@ module Spree
     module VendorsControllerDecorator
       def self.prepended(base)
         base.before_action :build_logo, only: %i[create update]
+        base.before_action :contruct_nearby_place, only: %i[create update]
       end
 
       private
+
+      def contruct_nearby_place
+        nearby_places_attributes = {}
+        permitted_resource_params[:nearby_places_attributes].each do |id, vendor_place_attributes|
+          nearby_places_attributes[id] = vendor_place_attributes.except(:selected) if
+          vendor_place_attributes[:selected] == 'true'
+        end
+        permitted_resource_params[:nearby_places_attributes] = nearby_places_attributes
+      end
 
       def build_logo
         return unless permitted_resource_params[:logo]
--- a/app/interactors/spree_cm_commissioner/nearby_place_distance_finder.rb
+++ b/app/interactors/spree_cm_commissioner/nearby_place_distance_finder.rb
@@ -1,0 +1,27 @@
+module SpreeCmCommissioner
+  class NearbyPlaceDistanceFinder < BaseInteractor
+    delegate :vendor, :place, to: :context
+
+    def call
+      context.distance = get_distance([vendor.lat, vendor.lon], [place.lat, place.lon])
+    end
+
+    protected
+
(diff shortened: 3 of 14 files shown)


app/controllers/spree/admin/nearby_places_controller.rb
```diff
@@ -4,17 +4,18 @@ class NearbyPlacesController < Spree::Admin::ResourceController
...
       def new
-        SpreeCmCommissioner::NearbyPlaceCreator.call(params: permitted_resource_params, vendor: @vendor)
+        @nearby_places = @vendor.nearby_places
       end
 
       private
 
       def load_vendor
         @vendor ||= Spree::Vendor.find_by(slug: params[:vendor_id])
+        @object ||= @vendor
```
kimsrung, line 18: to save this, we can just have @object then in our form @vendor => @object.

app/interactors/spree_cm_commissioner/nearby_places_loader.rb
```diff
@@ -0,0 +1,40 @@
...
+        NearbyPlaceDistanceFinder.call(vendor: vendor, place: place, index: index)
+      end
+    end
+
+    def google_map_key
+      @google_map_key ||= ENV.fetch('GOOGLE_MAP_KEY')
+    end
+
+    def request
+      url = URI("https://maps.googleapis.com/maps/api/place/nearbysearch/json?location=#{vendor.lat}%2C#{vendor.lon}&radius=5000&key=#{google_map_key}")
```
kimsrung, line 31: maybe we can create a variable called GOOGLE_MAP_API_URL="https://maps.googleapis.com/maps/api"

later if you have another feature like find_place, we can reuse that constant from env

config/routes.rb
```diff
@@ -30,6 +30,8 @@
       end
       resources :nearby_places do
         collection do
+          patch :update
+          patch :update_positions
           post :update_positions
```
kimsrung, line 35: you can remove post :update_positions . it's not necessary after adding patch :update_position. Please help to verify

app/interactors/spree_cm_commissioner/nearby_places_loader.rb
```diff
@@ -0,0 +1,41 @@
+module SpreeCmCommissioner
+  class NearbyPlacesLoader < BaseInteractor
+    delegate :vendor, to: :context
+
+    GOOGLE_MAP_API_URL = 'https://maps.googleapis.com/maps/api'.freeze
```
kimsrung, line 5: I'd prefer this in .env . what would you think?

app/controllers/spree/admin/nearby_places_controller.rb
```diff
@@ -25,20 +25,22 @@ def object_name
...
-
       def collection_url(options = {})
         admin_vendor_nearby_places_url(options)
       end
 
       def permitted_resource_params
-        params.permit(nearby_places: %i[name reference lat lon vicinity icon url rating formatted_address address_components])
+        nearby_places = []
+        promoted_nearby_place_ids = params['vendor']['promoted_nearby_place_ids']
+        promoted_nearby_place_ids = promoted_nearby_place_ids.map(&:to_i)
```
kimsrung, line 35: promoted_nearby_place_ids = params['vendor']['promoted_nearby_place_ids'].map(&:to_i)

maybe we can just assign 1 time here. Assigning multiple time with the same name is little bit confusing since it's not a long method

app/interactors/spree_cm_commissioner/nearby_places_loader.rb
panhachom: since there is no .env file  in commissioner, I think it is much easy to call the variable here

app/interactors/spree_cm_commissioner/nearby_places_loader.rb
```diff
@@ -0,0 +1,41 @@
...
+    def call
+      json_response = request
+
+      json_response['results'].each_with_index do |json, index|
```
channainfo, line 9: We should not save the place from search but to save only those were selected

config/routes.rb
```diff
@@ -30,6 +30,7 @@
       end
       resources :nearby_places do
         collection do
+          patch :update
```
channainfo, line 33: remove this. the **update** is a built in

channainfo (changes requested):

app/interactors/spree_cm_commissioner/nearby_places_loader.rb
kimsrung: what you have done for this ENV.fetch('GOOGLE_MAP_KEY'), you could apply for url as well

theachoem: It is a constant variable, i don't think putting it in .env is a good idea. If we want to reuse, we can access from here instead.

kimsrung: so if you have another service for fetching google map api for place search or auto complete, are we going to repeat that URL in each class? 

I was just thinking ahead but if we don't have that feature, we can just add constant there

theachoem: We can access via sth like this: NearbyPlacesLoader::GOOGLE_MAP_API_URL or put it in another alternative class and access like so. I'm sure most gem do this instead of ask user to put in env.

kimsrung: if so, we can put in initializers with google_map.rb

theachoem: Yeah that's better

config/routes.rb
panhachom: Screenshot from 2023-03-03 15-07-20

without set this it will trigger this error

theachoem:
@panhachom change commit name to this instead: https://github.com/channainfo/commissioner/issues/182

PR #226, opened 2023-03-10
close #225 add pricing model by extend with promotion

### LINKED ISSUE #225: Price Rule feature by extend Spree::Promotion
## How?
By default, when apply a promotion, it activates the promotion with the adjustable object & creates adjustments for them.

Current adjustable objects are:
- Order
- LineItem
- Shipment

To turn Promotion into a PriceRule, we have to make sure Promotion/PriceRule can activate to vendor, product, or variant. To do so, vendor, product,... have to be adjustable.

## Convert to adjustable

DB Diagram below demonstrates how to convert vendor, product, ... into adjustable objects.

> During implementation, we removed the required of order_id in Adjustment & also remove order_id from ListingPrice.

image

## References
- https://github.com/channainfo/commissioner/issues/223

theachoem:
## Scope

- [x] Add Interactor: _vendor listing price activator_:
  - Add vendor listing prices from from_date...to_date
  - Adjust eligible promotions to those vendor listing prices
- [x] Add Action: listing price adjustment action
- [x] Add Rule: fixed date rule
   - Eligible when date between rule _start date_ to _start date + length - 1_. 
- [x] Override validation of adjustment to not required order when its adjustable is `ListingPrice`.

## References
- Detail why & diagram: 
  https://github.com/channainfo/commissioner/issues/225

### DIFF
--- a/app/interactors/spree_cm_commissioner/vendor_price_rules_activator.rb
+++ b/app/interactors/spree_cm_commissioner/vendor_price_rules_activator.rb
@@ -1,0 +1,49 @@
+module SpreeCmCommissioner
+  class VendorPriceRulesActivator < BaseInteractor
+    delegate :from_date, :to_date, :vendor_id, to: :context
+
+    def call
+      load_vendor_listing_prices
+      adjust_vendor_listing_prices
+    end
+
+    # create listing prices for vendor on specific date
+    def load_vendor_listing_prices
+      context.vendor_listing_prices = (from_date..to_date).map do |date|
+        SpreeCmCommissioner::ListingPrice.first_or_initialize(
+          price_source: vendor,
+          date: date,
+          currency: Spree::Store.default.default_currency
+        ) do |vendor_listing_price|
+          vendor_listing_price.price = vendor.max_price
+        end
+      end
+    end
+
+    def adjust_vendor_listing_prices
+      context.vendor_listing_prices.each do |vendor_listing_price|
+        activate_pricing_models_to(vendor_listing_price)
+      end
+    end
+
+    def activate_pricing_models_to(modelable)
+      vendor_price_rules.each do |rule|
+        next unless eligible?(rule, modelable)
+
+        rule.pricing_model.activate(adjustable: modelable)
+      end
+    end
+
+    def eligible?(rule, modelable)
+      rule.applicable?(modelable) && rule.eligible?(modelable, {})
+    end
+
+    def vendor_price_rules
+      context.vendor_price_rules ||= vendor.price_rules
+    end
+
+    def vendor
+      context.vendor ||= Spree::Vendor.find(vendor_id)
+    end
+  end
+end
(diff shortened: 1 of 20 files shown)


app/interactors/spree_cm_commissioner/vendor_price_rules_activator.rb
```diff
@@ -0,0 +1,53 @@
...
+        rule.promotion.activate(adjustable: promotable, order: order)
+      end
+    end
+
+    def eligible?(rule, promotable)
+      rule.applicable?(promotable) && rule.eligible?(promotable, {})
+    end
+
+    def vendor_price_rules
+      context.vendor_price_rules ||= Spree::PromotionRule.where(vendor_id: vendor_id)
```
channainfo, line 46: Let's query from PriceRule instead of PromotionRule
```
context.vendor.price_rules
```

app/models/spree_cm_commissioner/promotion_rule_decorator.rb
```diff
@@ -0,0 +1,11 @@
+module SpreeCmCommissioner
+  module PromotionRuleDecorator
+    def self.prepended(base)
+      base.belongs_to :vendor, class_name: 'Spree::Vendor', dependent: :destroy
```
channainfo, line 4: Add this to PriceRule instead and let's not touch PromotionRule yet.

db/migrate/20230310100611_create_spree_cm_commissioner_listing_prices.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.timestamps
+    end
+
+    add_index :cm_listing_prices, [
+      :price_source_type,
+      :price_source_id,
+      :order_id,
+      :date,
+      :currency
+    ], unique: true, name: 'index_cm_listing_prices_uniqueness', if_not_exists: true
```
channainfo, line 28: it should be enough with just 3 fields for the index.
```
:price_source_type, :price_source_id, :order_id,
```

app/interactors/spree_cm_commissioner/vendor_price_rules_activator.rb
```diff
@@ -0,0 +1,53 @@
+module SpreeCmCommissioner
+  class VendorPriceRulesActivator < BaseInteractor
+    delegate :from_date, :to_date, :vendor_id, :order, to: :context
```
channainfo, line 3: Because PriceRule is done via product listing, let's try to ignore the **order** field for now to avoid confusion.

channainfo (changes requested):

db/migrate/20230310100611_create_spree_cm_commissioner_listing_prices.rb
theachoem: > it should be enough with just 3 fields for the index.
> 
> ```
> :price_source_type, :price_source_id, :order_id,
> ```

We want it to be unique base on these columns. 
- Price source is vital for search
- Date. 2021-12-21
- Currency, sometimes price different base on currency, so I think we should allow user to create new if currently is different.

PR #227, opened 2023-03-11
close #105 add sort search option type list page

### LINKED ISSUE #105: Sort & search option type list page
Would be cool if allow user to search by name, sort by kind, name, etc.

image

Regene27:
Screenshot from 2023-03-11 13-55-17

### DIFF
--- a/app/controllers/spree/admin/option_types_controller_decorator.rb
+++ b/app/controllers/spree/admin/option_types_controller_decorator.rb
@@ -1,0 +1,16 @@
+module Spree
+  module Admin
+    module OptionTypesControllerDecorator
+      def collection_finder
+        @collection_finder = params[:kind].present? ? Spree::OptionType.where(kind: params[:kind]) : Spree::OptionType.all
+      end
+
+      def collection
+        @search = collection_finder.ransack(params[:q])
+        @collection = @search.result
+      end
+    end
+  end
+end
+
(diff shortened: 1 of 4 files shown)


channainfo:
@Regene27 Fix the quality code failure

app/controllers/spree/admin/option_types_controller_decorator.rb
```diff
@@ -0,0 +1,15 @@
+module Spree
+  module Admin
+    module OptionTypesControllerDecorator
+      def collection
+        @collection ||= Spree::OptionType.where(kind: params[:kind]) if params[:kind].present?
+        @collection ||= Spree::OptionType.all
+
```
kimsrung, line 7: we may zip this login to a method called collection_finder

```
def collection_finder
   @collection_finder = params[:kind].present? ? Spree::OptionType.where(kind: params[:kind]) : Spree::OptionType.all
end
```

app/controllers/spree/admin/option_types_controller_decorator.rb
```diff
@@ -0,0 +1,15 @@
...
+        @search = @collection.ransack(params[:q])
+        @collection = @search.result
```
kimsrung, line 9: we don't need @search, we can just assign 
```
@collection = @ collection_finder.ransack(params[:q]).result
```

kimsrung (approved):

PR #261, opened 2023-03-23
Close #260 generate sequence number of invoice

### LINKED ISSUE #260: Generate Sequence Number for Invoice

panhachom:
Screenshot 2023-03-23 at 2 29 23 PM

### DIFF
--- a/app/interactors/spree_cm_commissioner/invoice_number_creator.rb
+++ b/app/interactors/spree_cm_commissioner/invoice_number_creator.rb
@@ -1,0 +1,48 @@
+module SpreeCmCommissioner
+  class InvoiceNumberCreator < BaseInteractor
+    delegate :order, to: :context
+
+    def call
+      load_invoice_number
+      create_invoice
+    end
+
+    # 0412 + 000001
+    def load_invoice_number
+      load_prefix
+      load_invoices_count
+
+      context.invoice_number = context.prefix + rjust_number(context.invoices_count)
+    end
+
+    # 04-12-2023 -> 0412
+    def load_prefix
+      context.prefix = order.line_items.first.from_date.strftime('%m%y')
+    end
+
+    def load_invoices_count
+      from = order.line_items.first.from_date.beginning_of_month
+      to = order.line_items.first.from_date.end_of_month
+
+      context.invoices_count = SpreeCmCommissioner::Invoice.where(
+        vendor: order.line_items.first.vendor,
+        date: from..to
+      ).size
+    end
+
+    def create_invoice
+      context.invoice = SpreeCmCommissioner::Invoice.where(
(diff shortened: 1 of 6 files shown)


app/interactors/spree_cm_commissioner/invoice_number_generator.rb
```diff
@@ -0,0 +1,48 @@
...
+      context.invoice_number = context.prefix + rjust_number(context.invoices_count)
+    end
+
+    def load_invoices_count
+      from = order.line_items.first.from_date.beginning_of_month
+      to = order.line_items.first.from_date.end_of_month
+
+      context.invoices_count = SpreeCmCommissioner::Invoice.where(
+        vendor: order.line_items.first.vendor,
+        order: order,
```
channainfo, line 29: Let's remove the order

spec/interactors/spree_cm_commissioner/invoice_number_generator_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper'
+
+RSpec.describe SpreeCmCommissioner::InvoiceNumberGenerator do
+
+
+  let(:product) { create(:cm_product) }
+  let(:line_item) { create(:line_item, variant: product.master, from_date: '2023-06-07') }
+  let(:order1) { create(:order, line_items: [line_item]) }
+
```
channainfo, line 9: let's add another order.
Already test if the invoice already exist it should not create invoice again but return the existing invoice instead.

app/interactors/spree_cm_commissioner/invoice_number_generator.rb
```diff
@@ -0,0 +1,48 @@
+module SpreeCmCommissioner
+  class InvoiceNumberGenerator < BaseInteractor
```
channainfo, line 2: it seems like the purpose of this class is to create an invoice:  InvoiceCreator instead of generating an invoice number: InvoiceNumberGenerator

app/interactors/spree_cm_commissioner/invoice_number_generator.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      context.invoices_count = SpreeCmCommissioner::Invoice.where(
+        vendor: order.line_items.first.vendor,
+        order: order,
+        date: from..to
+      ).size
+    end
+
+    def create_invoice
+      context.invoice = SpreeCmCommissioner::Invoice.create!(
```
channainfo, line 35: we should not create a new invoice if the invoice for the order already existed ( not in canceled state).

channainfo (changes requested):

theachoem (approved):

PR #263, opened 2023-03-23
close #262 move index order of subscriptions to member action

### LINKED ISSUE #262: Move index orders of subscription to member action

theachoem:
- UI still same, just move order into member action, instead of its code MVC. 
- We will have another order controller for all orders of vendor instead.

image


app/controllers/spree/admin/merchant/subscriptions_controller.rb
```diff
@@ -5,6 +5,13 @@ class SubscriptionsController < Spree::Admin::Merchant::BaseController
         before_action :load_customer
         before_action :load_subscription, if: -> { member_action? }
 
+        def orders
```
channainfo, line 8: This will confuse and it is an anti pattern. Let's keep orders on its controller.

config/routes.rb
```diff
@@ -14,9 +14,12 @@
       scope ':vendor_id' do
         resources :customers do
           resources :subscriptions do
-            resources :orders
```
channainfo, line 17: technically it is the same resources :orders, only: %i[index]
let's not use orders as an action.

channainfo (changes requested):

config/routes.rb
theachoem: After this, we will have another order controller. Which under merchant/orders instead.

We move this to member as it sub of subscriptions. merchant/customers/subscriptions/orders. (It UI also under subscription tab)

theachoem: Alternative is user/orders

| | |
| - | - |
| image | image |

PR #266, opened 2023-03-24
Close #253 order invoice view

### LINKED ISSUE #253: Order Invoice View

panhachom:
Screenshot 2023-03-24 at 5 56 47 PM


config/routes.rb
```diff
@@ -14,7 +14,9 @@
       scope ':vendor_id' do
         resources :customers do
           resources :subscriptions do
-            resources :orders
+            resources :orders do
+              resources :invoice, only: [:index]
```
channainfo, line 18: use resource :invoices, only: %i[ create ].

InvoiceCreator should take care of the existence. If the invoice does not exist, it will create otherwise it returns the existing.

channainfo (changes requested):

PR #268, opened 2023-03-24
Close #267 enhance service calendar

### LINKED ISSUE #267: Enhance service calendar
- order calendar with latest first
- fix reason field not working using text_field instead
- change inclusion/exclusion to open/close

kimsrung:


channainfo:
I like the fact that you display with label: Open/Close yet internal we still use inclusion/exclusion. The original terms are more standardize. Let's keep the internal values as they are for now.

kimsrung:
Sorry bong. I was confused about your comments. Anyways, there are some points that I need to improve. I will update that part only

PR #277, opened 2023-03-28
Close #245 allow searching for a certain interval of time

### LINKED ISSUE #245: Allow searching for a certain interval of time
Calculating hotels with their availabilities is an extensive process. If the search range is huge the system would not be quick enough to respond. we might not allow searching for hotels for more than a week stay however this number of day should be configurable via Spree config / ENV

- Add validation layer for search API
- Render errors if intended params passed to the search failed
- Use Dry validation  https://github.com/dry-rb/dry-validation

kimsrung:
ADDED: another validation for search only bigger than current date.


app/interactors/spree_cm_commissioner/accommodation_search_detail.rb
```diff
@@ -68,6 +68,18 @@ def prepare_params
     def passenger_options
       SpreeCmCommissioner::PassengerOption.new(adult: params[:adult], children: params[:children], room_qty: params[:room_qty])
     end
+
+    def validate_search
+      context.fail!(error: contract_error) if contract.failure?
+    end
+
+    def contract
+      @contract ||= SpreeCmCommissioner::AccommodationSearchContract.new.call(from_date: from_date, to_date: to_date)
```
channainfo, line 77: @kimsrung let's remove this and construct the validator at the controller level instead. let's create contracts ( validation schemas ) in its folder. You might find this https://github.com/bookmebus/vshop168/commit/e0c8648cac0d09026079708837064cae51152c0b useful

channainfo (changes requested):

config/initializers/accommodation_search.rb
```diff
@@ -0,0 +1 @@
+ACCOMMODATION_MAX_STAY_DAYS = 10
```
channainfo, line 1: let's extract this to env. If not defined default to '10'  ENV.fetch('ACCOMMODATION_MAX_STAY_DAYS', '10'). Add this env to README as well

channainfo (changes requested):

spec/request/spree/api/v2/accommodations_spec.rb
```diff
@@ -219,4 +252,12 @@ def book_room(order, hotel: , price: 100, quantity: , from_date:, to_date:)
     room = hotel.variants.first
     order.line_items.create(vendor_id: hotel.id, price: price, quantity: quantity, from_date: from_date, to_date: to_date, variant_id: room.id)
   end
+
+  def day_to_s(day)
+    day.strftime("%F")
+  end
+
+  def max_stay_days
+    ENV.fetch('ACCOMMODATION_MAX_STAY_DAYS', 10)
```
channainfo, line 261: ENV.fetch returns a string. If we use default as int there might be an unstable outcome. 

```
 ENV.fetch('ACCOMMODATION_MAX_STAY_DAYS', '10').to_i
```

channainfo (changes requested):

PR #297, opened 2023-04-01
close #292 acl adding role authorization

### LINKED ISSUE #292: ACL Role Authorization

Regene27:


spec/controllers/concerns/spree_cm_commissioner/admin/role_authorization_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+class FakeController < ApplicationController
+  include SpreeCmCommissioner::Admin::RoleAuthorization
```
channainfo, line 4: Remove this. Use the parent controller of the merchant instead.

app/controllers/spree/admin/base_controller_decorator.rb
```diff
@@ -0,0 +1,11 @@
+module Spree
+  module Admin
+    module BaseControllerDecorator
+      def self.prepended(base)
+        base.include SpreeCmCommissioner::Admin::RoleAuthorization
```
channainfo, line 5: Narrow this to parent of the merchant.

channainfo (changes requested):

PR #411, opened 2023-06-12
Close #407 add start date and end date to line item

### LINKED ISSUE #407: Add start date & end date to line item UI
image

leak17:
Screenshot from 2023-06-13 14-26-00


app/overrides/spree/admin/orders/_line_items/_line_items.html.erb.deface
```diff
@@ -0,0 +1,49 @@
+<!-- replace ".table-responsive" -->
```
theachoem, line 1: Replace via data-hook instead (of line 3).

```
<!-- replace "[data-hook='line-items']" -->
```

Issue #449, opened 2023-07-03
Add home screen dynamic background image


theachoem:
# Scope
- Add `SpreeCmCommissioner::HomePageBackground` that includes `HomepageBackgroundAppImage` & `HomepageBackgroundWebImage`.
  Similar implementation: app/models/spree_cm_commissioner/homepage_banner.rb

- Make UI in admin to upload that image
  File to check:
  - app/controllers/spree/admin/homepage_banner_controller.rb

- Make API respond to that data to user:
  File to check:
   - app/controllers/spree/api/v2/storefront/homepage_data_controller.rb
  - app/services/spree_cm_commissioner/homepage_data_loader.rb
  - app/serializers/spree_cm_commissioner/v2/storefront/homepage_data_serializer.rb

<img src="https://github.com/channainfo/commissioner/assets/29684683/32dad8a4-03b8-4036-a365-0cf8ce670dde" width="350px" />

PR #462, opened 2023-07-10
WIP #449 add home screen dynamic background image

leak17:


app/models/spree_cm_commissioner/homepage_background.rb
```diff
@@ -15,5 +15,15 @@ def toggle_status!
 
       save!
     end
+
+    def generate_url
```
theachoem, line 19: We don't need this.

theachoem: You should recheck the related serializer instead.

leak17:
Ok bong, Noted

PR #513, opened 2023-08-11
Close #512 add an active field true or false in the homepage

### LINKED ISSUE #512: Add an active field, true or false, in the homepage

leak17:
https://github.com/channainfo/commissioner/assets/115885247/4695c520-e07b-4447-ade6-49f66b77cd60

### DIFF
--- a/app/views/spree/admin/homepage_background/_form.html.erb
+++ b/app/views/spree/admin/homepage_background/_form.html.erb
@@ -13,4 +13,8 @@
     <%= f.file_field :web_image %>
     <%= image_tag main_app.url_for(@object.web_image.url(:small)) if @object.web_image %>
   <% end %>
+  <%= f.field_container :active do %>
+    <%= f.label :active %>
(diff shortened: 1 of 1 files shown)


channainfo:
@leak17 add demo when you change the value if it saves correctly

app/views/spree/admin/homepage_background/edit.html.erb
```diff
@@ -5,6 +5,10 @@
 <%= form_with model: @object, url: { action: 'update' }, html: { multipart: true } do |f| %>
   <fieldset>
     <%= render partial: 'form', locals: { f: f } %>
+     <div class="field">
+      <%= f.label :active %>
```
theachoem, line 9: Put this in form.

theachoem (changes requested):

leak17:
> @leak17 add demo when you change the value if it saves correctly

noted bong, it has some issues; let me correct it and put the demo again.

PR #521, opened 2023-08-22
close #520 added a qrcode payment field to vendor

### LINKED ISSUE #520: add_qrcode_field

tykeaboyloy:
-add vendor edit section to billing
image

-add qrcode field to vendor asset
-use the qrcode in invoice for payment
image

### DIFF
--- a/app/models/spree_cm_commissioner/vendor_qrcode.rb
+++ b/app/models/spree_cm_commissioner/vendor_qrcode.rb
@@ -1,0 +1,12 @@
+module SpreeCmCommissioner
+  class VendorQrcode < Asset
+    # 16x9
+    def asset_styles
+      {
+        mini: '160x90>',
+        small: '480x270>',
+        medium: '960x540>'
+      }
+    end
+  end
+end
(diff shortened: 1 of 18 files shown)


app/models/spree_cm_commissioner/vendor_qrcode.rb
```diff
@@ -0,0 +1,12 @@
+module SpreeCmCommissioner
+  class VendorQrcode < Asset
```
channainfo, line 2: QR code is not a form of representing URL in a machine-readable way. You can use ruby or js lib to generate the QRCode by taking an URL as input. No need to store it as an asset.

channainfo (changes requested):

app/models/spree_cm_commissioner/vendor_qrcode.rb
tykeaboyloy: Bong if we store the url and use it to generate the QRCode ourselves, the generated QRCode can only be scanned through third party scanner, but cannot be scanned using bank app (aba, acleda).

channainfo:
@tykeaboyloy let's rename it to payment_qrcode to be not confused with a normal qr.

PR #543, opened 2023-09-02
 close #542 rename vendor state to default state to avoid confusion, add to serializer

### LINKED ISSUE #542: Improve api - add state name and fix option value in case on assets found

theachoem:
1. In vendor, rename vendor.state_id to vendor.defaut_state_id to avoid with vendor.state (active, blocked,...)
2. Add default_state to accommodation serializer
3. Add state to stock location serializer
4. Fix bug on option value serializer when no asset or asset failed to load

### DIFF
--- a/app/serializers/spree/v2/storefront/accommodation_serializer.rb
+++ b/app/serializers/spree/v2/storefront/accommodation_serializer.rb
@@ -7,7 +7,7 @@ module Spree
         # because its model still Spree::Vendor
         set_type :vendor
 
-        has_one :state
+        has_one :default_state, serializer: :state
 
         attributes :total_inventory, :service_availabilities
 
--- a/app/models/spree_cm_commissioner/state_decorator.rb
+++ b/app/models/spree_cm_commissioner/state_decorator.rb
@@ -1,7 +1,7 @@
 module SpreeCmCommissioner
   module StateDecorator
     def self.prepended(base)
-      base.has_many :vendors
+      base.has_many :vendors, foreign_key: 'default_state_id', class_name: 'Spree::Vendor', inverse_of: :default_state, dependent: :nullify
 
       def update_total_inventory
         update(total_inventory: vendors.pluck(:total_inventory).sum)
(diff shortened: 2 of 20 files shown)


app/serializers/spree/v2/storefront/accommodation_serializer.rb
```diff
@@ -7,14 +7,18 @@ class AccommodationSerializer < VendorSerializer
...
 
-        has_one :state
-
         attributes :total_inventory, :service_availabilities
 
         attribute :total_booking do |vendor|
           vendor.respond_to?(:total_booking) ? vendor.total_booking : 0
         end
 
+        attribute :state_name do |vendor|
```
channainfo, line 16: @theachoem let's not do this in the serializer. It can be dangerous to be called on the collection and then we end up with n+1 query and hard introspect.

channainfo: one way to get this is to have a callback in stock_location after_commit to create the vendor state if the state does not exist yet and do a migration

channainfo (changes requested):

app/serializers/spree/v2/storefront/accommodation_serializer.rb
```diff
@@ -7,14 +7,18 @@ class AccommodationSerializer < VendorSerializer
         # because its model still Spree::Vendor
         set_type :vendor
 
-        has_one :state
```
channainfo, line 10: let's keep this so that we know that the serializer might need to eager load the state.

channainfo (changes requested):

app/serializers/spree/v2/storefront/accommodation_serializer.rb
theachoem: should we rename the state column bong? I do this because serialier confuse the vendor.state_id & vendor.state (active, pending, ..).

I think I should rename vendor.state to status instead.

PR #554, opened 2023-09-06
Close #553 order request notification api

### LINKED ISSUE #553: Order Request notification api

panhachom:


app/controllers/spree/api/v2/storefront/order_request_notifications_controller.rb
```diff
@@ -0,0 +1,39 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class OrderRequestNotificationsController < Spree::Api::V2::ResourceController
+          before_action :require_spree_current_user
+
+          def collection
+            spree_current_user.notifications.newest_first.request_notifications
```
channainfo, line 9: let's combine the newest_first.request_notifications into just one by calling the newest in the request_notifications directly to avoid many nested scopes for the public API call.

spree_current_user.notifications.top_3_request_notifications

channainfo (changes requested):

PR #557, opened 2023-09-07
Close #547 transit branch setup

### LINKED ISSUE #547: Transit Branch Setup

Regene27:
Screenshot from 2023-09-11 17-14-25
image

### DIFF
--- a/app/controllers/spree/transit/branches_controller.rb
+++ b/app/controllers/spree/transit/branches_controller.rb
@@ -1,0 +1,35 @@
+module Spree
+  module Transit
+    class BranchesController < Spree::Transit::BaseController
+
+      def new
+        @branch = SpreeCmCommissioner::Branch.new
+        super
+      end
+
+      def collection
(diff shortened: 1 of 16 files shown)


channainfo:
@Regene27 looks good except that the branch belongs to a vendor. You should not be able to create a branch without a vendor.

channainfo:
@Regene27 resolve conflicts

app/controllers/spree/transit/branches_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Spree
+  module Transit
+    class BranchesController < Spree::Transit::BaseController
+      before_action :load_branch
+
+      def load_branch
```
channainfo, line 6: you don't need this. the branches can be accessed simply using:

```
current_vendor.branches
```
it should be available in the view. No need for extra boilerplate code.

channainfo (changes requested):

PR #573, opened 2023-09-13
Close #566 add branch filtering

### LINKED ISSUE #566: add branch filtering

Regene27:
Screenshot from 2023-09-13 15-37-17
Screenshot from 2023-09-13 15-36-58

### DIFF
--- a/app/controllers/spree/transit/branches_controller.rb
+++ b/app/controllers/spree/transit/branches_controller.rb
@@ -8,10 +8,9 @@ module Spree
 
       def collection
         return @collection if defined?(@collection)
-
         current_vendor.branches
 
-        @search = current_vendor.branches.ransack(params[:q])
+        @search = current_vendor.branches.includes(:state).ransack(params[:q])
         @collection = @search.result
       end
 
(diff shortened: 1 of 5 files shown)


app/controllers/spree/transit/branches_controller.rb
```diff
@@ -12,7 +12,7 @@ def collection
         current_vendor.branches
 
         @search = current_vendor.branches.ransack(params[:q])
```
channainfo, line 14: eager load the state here: includes(:state) because you access the branch.state in the view

channainfo (changes requested):

app/controllers/spree/transit/branches_controller.rb
```diff
@@ -12,7 +12,7 @@ def collection
...
-        @collection = @search.result
+        @collection = @search.result.page(page).per(per_page)
```
channainfo, line 15: this is enough @search.result.page(page) no need to explicitly call the **per**

channainfo (changes requested):

PR #617, opened 2023-09-24
WIP add email to user profile api

vaneath:


theachoem:
Close as we can just use update API

PR #624, opened 2023-09-26
close #623 send pin code to telegram for debugging purpose

### LINKED ISSUE #623: Send pin code to telegram for debugging purpose

theachoem:
- Send pin code to telegram channel
- Disabled on production

image


app/interactors/spree_cm_commissioner/pin_code_sender.rb
```diff
@@ -8,6 +8,8 @@ def call
       else
         send_email
       end
+
+      send_telegram_debug_pin_code if Rails.env.production?
```
channainfo, line 12: I think this may be useful. Let's set an ENV value to turn this on/off rather than Rails.env. Rails.env is problematic and should be used to check the lib available level only.

channainfo (changes requested):

PR #627, opened 2023-09-27
close #626 fix cannot sum vendor total inventory to state

### LINKED ISSUE #626: Can't sum state total inventory

theachoem:
Cannot sum because some inventory unit is nil

### DIFF
--- a/app/models/spree_cm_commissioner/state_decorator.rb
+++ b/app/models/spree_cm_commissioner/state_decorator.rb
@@ -4,7 +4,7 @@ module SpreeCmCommissioner
       base.has_many :vendors, foreign_key: 'default_state_id', class_name: 'Spree::Vendor', inverse_of: :default_state, dependent: :nullify
 
       def update_total_inventory
-        update(total_inventory: vendors.pluck(:total_inventory).sum)
+        update(total_inventory: vendors.pluck(:total_inventory).compact.sum)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/models/spree_cm_commissioner/state_decorator.rb
```diff
@@ -4,7 +4,7 @@ def self.prepended(base)
       base.has_many :vendors, foreign_key: 'default_state_id', class_name: 'Spree::Vendor', inverse_of: :default_state, dependent: :nullify
 
       def update_total_inventory
-        update(total_inventory: vendors.pluck(:total_inventory).sum)
+        update(total_inventory: vendors.with_inventory.pluck(:total_inventory).sum)
```
channainfo, line 7: @theachoem this may cause n+1 query. How about converting all the total_invtentory to a number, thus nil become 0 and ignores the where query. we can also set a default value for total inventory to be cero.

channainfo (changes requested):

spec/models/spree/state_spec.rb
```diff
@@ -16,4 +16,16 @@
...
+  describe '#update_total_inventory' do
+    it 'do' do
+      state = create(:state)
+
+      vendor1 = build(:cm_vendor, total_inventory: 1)
+      vendor2 = build(:cm_vendor, total_inventory: nil)
+
+      state.vendors = [vendor1, vendor2]
+      state.update_total_inventory
+    end
```
channainfo, line 29: add spec that state#total_inventory is eq to 1

channainfo (changes requested):

PR #629, opened 2023-09-27
Close #628 line-items api 

### LINKED ISSUE #628: Line items api

panhachom:


app/models/concerns/spree_cm_commissioner/order_filter_scopes.rb
```diff
@@ -0,0 +1,35 @@
+module SpreeCmCommissioner
+  module OrderFilterScopes
+    extend ActiveSupport::Concern
+
+    included do
+      scope :filter_by_event, lambda { |event|
+        case event
+        when 'upcoming'
+          complete.joins(:line_items).where('spree_line_items.to_date < ?', Time.current)
```
channainfo, line 9: upcoming: to_date >= Date.current
completed to_date < Date.current

app/models/concerns/spree_cm_commissioner/order_filter_scopes.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      scope :filter_by_request_state, lambda {
+                                        where(state: :complete, payment_state: :paid)
+                                          .where.not(request_state: nil)
+                                          .order(created_at: :desc)
+                                      }
+
+      def event_status
+        if complete?
+          return 'complete' if line_items.all? { |line_item| line_item.to_date && line_item.to_date > Time.current }
```
channainfo, line 27: use Date.current and with the. >=

app/models/concerns/spree_cm_commissioner/order_filter_scopes.rb
channainfo: Upcoming and complete should be filtered on the line_items. Let's simplify by displaying the line items instead of the orders. Having orders in the upcoming and complete are confusing and complicated.

channainfo (changes requested):

app/controllers/spree/api/v2/storefront/line_items_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class LineItemsController < ::Spree::Api::V2::ResourceController
+          before_action :require_spree_current_user
+
+          def collection
+            spree_current_user.line_items.filter_by_event(params[:event])
```
theachoem, line 9: Allow sort here, implement can be similar to orders_controller which support multi sorts.

theachoem: one must have field is :to_date, user can sort latest events

theachoem: image

https://api.spreecommerce.org/docs/api-v2/973a8ce6cadad-list-all-products

app/serializers/spree/v2/storefront/line_item_serializer_decorator.rb
```diff
@@ -4,6 +4,10 @@ module Storefront
       module LineItemSerializerDecorator
         def self.prepended(base)
           base.attributes :from_date, :to_date, :vendor_id, :need_confirmation, :product_type
+          base.attribute :order_number do |line_item|
+            line_item.order.number
```
channainfo, line 8: Create a new line_item serializer and use  belongs_to :order instead, this way we will not endup using n+1 query. When we want the order ( number ) we can invoke includes(:order)

channainfo (changes requested):

PR #631, opened 2023-09-27
close #630 add application controller decorator to ActionController::API

### LINKED ISSUE #630: Add ApplicationControllerDecorator logics to ActionController:API

theachoem:


app/controllers/spree_cm_commissioner/action_controller/api_decorator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      # Annonymous block: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/BlockForwarding
+      def set_writing_role(&)
+        ActiveRecord::Base.connected_to(role: :writing, &)
+      end
+    end
+  end
+end
+
+unless ActionController::API.included_modules.include?(SpreeCmCommissioner::ActionController::ApiDecorator)
```
channainfo, line 16: Can you have it to a specific controller?  If we do a writing role to the root controller, reading role will never be used

channainfo (changes requested):

app/controllers/spree_cm_commissioner/action_controller/api_decorator.rb
theachoem: We don't call it yet bong. If not put it, would raise error on child api controller that call that method.

PR #656, opened 2023-10-04
Close #655 enhance vendor query response for JsonLD

### LINKED ISSUE #655: enhance vendor query response for JsonLD
adding image_url attributes to vendor serializer for JsonLD

kimsrung:
image

### DIFF
--- a/app/controllers/spree/api/json_ld/vendors_controller.rb
+++ b/app/controllers/spree/api/json_ld/vendors_controller.rb
@@ -1,0 +1,15 @@
+module Spree
+  module Api
+    module JsonLd
+      class VendorsController < ::Spree::Api::V2::Storefront::VendorsController
+        def resource_serializer
+          SpreeCmCommissioner::JsonLd::VendorSerializer
+        end
+
+        def collection_serializer
+          SpreeCmCommissioner::JsonLd::VendorSerializer
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/spree/api/json_ld/vendors_controller.rb
+++ b/app/controllers/spree/api/json_ld/vendors_controller.rb
@@ -9,6 +9,10 @@ module Spree
         def collection_serializer
           SpreeCmCommissioner::JsonLd::VendorSerializer
         end
+
+        def scope
+          ::Spree::Vendor.active.includes(:image)
+        end
       end
     end
   end
--- a/app/models/spree_cm_commissioner/vendor_decorator.rb
+++ b/app/models/spree_cm_commissioner/vendor_decorator.rb
@@ -137,6 +137,14 @@ module SpreeCmCommissioner
     def generate_code
       self.code = (code.presence || name[0, 3].upcase)
     end
+
+    def full_address
+      address = stock_location.address1 || stock_location.address2
+      address += "  #{stock_location.city}" if stock_location.city.present?
+      state_name = Spree::State.find(stock_location.state_id)&.name if stock_location.state_id.present?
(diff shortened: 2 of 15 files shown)


app/models/spree_cm_commissioner/vendor_decorator.rb
```diff
@@ -137,6 +137,14 @@ def selected_option_value_vendors_ids
     def generate_code
       self.code = (code.presence || name[0, 3].upcase)
     end
+
+    def full_address
```
channainfo, line 141: This method is complicated and resource-consuming and it is easy to mislead for example, if we call this method with a list of vendors, it will cause a crazy query and be very resource-intensive.

If this info is important, let's just create a new field in the vendor and read from it instead.

channainfo (changes requested):

app/serializers/spree/v2/storefront/vendor_serializer_decorator.rb
```diff
@@ -17,6 +17,12 @@ def self.prepended(base)
           base.has_many :active_promotions, serializer: ::SpreeCmCommissioner::V2::Storefront::PromotionSerializer
 
           base.has_one :logo, serializer: ::SpreeCmCommissioner::V2::Storefront::AssetSerializer
+
+          base.attribute :image_url do |vendor|
+            image = vendor.image || vendor.logo
```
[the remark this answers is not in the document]
kimsrung: yes, we need that image for JSON-LD only. Sometimes, logo is blank. That's y I use image or logo.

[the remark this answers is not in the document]
channainfo: If you need a specific logic for JSON-LD, create a new serializer under jsonld instead. Take into account that adding a new model to a serializer will put a huge burden on others who use the same serializer.

channainfo (changes requested):

app/controllers/spree/api/json_ld/vendors_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+    module JsonLd
+      class VendorsController < ::Spree::Api::V2::Storefront::VendorsController
+        def resource_serializer
+          SpreeCmCommissioner::JsonLd::VendorSerializer
+        end
+
+        def collection_serializer
+          SpreeCmCommissioner::JsonLd::VendorSerializer
+        end
+      end
```
channainfo, line 12: we need to make sure that the collection is eager load the image includes(:image)

app/serializers/spree_cm_commissioner/json_ld/vendor_serializer.rb
```diff
@@ -0,0 +1,15 @@
+module SpreeCmCommissioner
+  module JsonLd
+    class VendorSerializer < ::Spree::V2::Storefront::VendorSerializer
+      set_type :vendor
+
+      attributes :image_url
+
```
channainfo, line 7: has_one :image

app/serializers/spree_cm_commissioner/json_ld/vendor_serializer.rb
```diff
@@ -0,0 +1,15 @@
+module SpreeCmCommissioner
+  module JsonLd
+    class VendorSerializer < ::Spree::V2::Storefront::VendorSerializer
+      set_type :vendor
+
+      attributes :image_url
```
channainfo, line 6: remove this, we use includes(:image) instead.

channainfo (changes requested):

PR #658, opened 2023-10-04
close #657 add event date to taxon

### LINKED ISSUE #657: add event date to Taxon

panhachom:
image
### Task Breakdown
- add to_date and from_date (datetime) to Spree::Taxon
- add validation where product can have only 1 event taxon 
- after line-item created, set line_item.from_date & to_date from taxon.from_date & to_date, so we can fetch upcoming in app.


theachoem (approved):
LGTM!

app/models/spree_cm_commissioner/taxon_decorator.rb
```diff
@@ -7,6 +7,11 @@ def self.prepended(base)
       base.has_one :app_banner, as: :viewable, dependent: :destroy, class_name: 'SpreeCmCommissioner::TaxonAppBanner'
 
       base.validates_associated :category_icon
+      base.scope :event, -> { where('permalink LIKE ?', 'events%') }
```
channainfo, line 10: when we add translation, the events permalink may be changed based on the language of translation. Let's create a new column for taxonomy and taxon called "kind" using enum. On the before save in taxon, we copy valid of kind from taxonomy to the taxon. 

possible enum values for kind: category: 0, cms: 1, event: 2

channainfo (changes requested):

PR #719, opened 2023-10-18
Close #718 improve public qr generator

### LINKED ISSUE #718: Improve public QR generator
Provide an ability to do a check to safeguard server resources.

- The QR data "#{order.number}-#{order.token}"
- At the server side, check the format of the QR data ( it should start with R, the token length ) before querying to make sure it is not overwhelming the DB server.
- Query by order.token to get the order.

kimsrung:


app/controllers/spree_cm_commissioner/qr_images_controller.rb
```diff
@@ -1,6 +1,10 @@
 class SpreeCmCommissioner::QrImagesController < ApplicationController
   def show
-    order = Spree::Order.find_by!(token: params[:id])
+    token = params[:id].match(/^R\d{9,}-([A-Za-z0-9_\-]+)$/)&.captures
+
+    return unless token
```
channainfo, line 5: if the token is not valid we can raise not found status code. We need to add a spec for this.

I think it would be better to have a method in the Order to find the order by qr_data. we test this method instead of the controller.

channainfo (changes requested):

PR #802, opened 2023-11-13
Close #769 discount maximum cap

### LINKED ISSUE #769: Discount maximum (CAP)
@channainfo

kimsrung:


app/models/spree_cm_commissioner/order_decorator.rb
```diff
@@ -39,6 +40,20 @@ def base.search_by_qr_data!(data)
...
     end
 
+    def adjust_discount_by_cap
+      promotion_adjustment = adjustments.promotion.eligible.first
+
+      return if promotion_adjustment.blank? # Check if the order has a promotion adjustment
+
+      promotion = promotion_adjustment.source.promotion
+
+      return unless promotion&.cap # Check if cap exist
```
channainfo, line 50: @kimsrung compare their abs value instead.

app/models/spree_cm_commissioner/order_updater_decorator.rb
```diff
@@ -5,6 +5,13 @@ def update_item_total
       order.item_total = line_items.sum(&:amount)
       update_order_total
     end
+
+    def update_adjustment_total_by_cap(cap)
+      update_adjustment_total
```
channainfo, line 10: let's merge this with the above because this method is dangerous it allows updating the total by a value (cap) without checking the condition and if it gets called anywhere else it will be incorrect.

channainfo (changes requested):

theachoem:
It is great for whole order adjustment, but It will not work with item adjustment Or date-specific item adjustment.

kimsrung:
> It is great for whole order adjustment, but It will not work with item adjustment Or date-specific item adjustment.
> 
> Another approach we can do is putting CAP inside calculable of Spree::Calculator instead.
> 
> `calculator is belongs_to :calculable`. Caculatable here can be CreateAdjustment, CreateItemAdjustments, or CreateDateSpecificItemAdjustments.
> 
> Then, we can override compute method or override it logic based on a specific calculator:
> 
> ```ruby
> def compute
>    return super unless calculable.defined?(:cap)
> 
>    [super, calculable.cap].min
> end
> ```
> 
> This will be a lot of change but please take it into consideration.
> 
> 

From what I know, we apply cap to whole order only. We would get final price and calculate. And this cap would be on promotion only. In case you apply the promotion without cap, everything is still the same. We can write spec to verify this.

app/models/spree_cm_commissioner/order_updater_decorator.rb
```diff
@@ -5,6 +5,23 @@ def update_item_total
...
+
+      promotion_adjustment = adjustments.promotion.eligible.first
+
+      return if promotion_adjustment.blank? # Check if the order has a promotion adjustment
+
+      promotion = promotion_adjustment.source.promotion
+
+      return unless promotion&.cap && order.adjustment_total.abs > promotion&.cap # Check if cap exist
+
+      order.adjustment_total = order.promo_total = promotion.cap * -1
```
channainfo, line 21: create a new var to represent promotion.cap * -1 and give it a meaningful name.

app/models/spree_cm_commissioner/calculator/percent_on_line_item_decorator.rb
```diff
@@ -0,0 +1,18 @@
+module SpreeCmCommissioner
+  module Calculator
+    module PercentOnLineItemDecorator
+      def compute(object)
```
channainfo, line 4: do you think we can find a universal place to override this method for all of the calculators instead of having to override one by one.

channainfo (changes requested):

app/models/spree_cm_commissioner/promotion/actions/create_date_specific_item_adjustments.rb
```diff
@@ -36,8 +36,11 @@ def compute_amount(line_item)
         def compute_line_item_amount(line_item)
           line_item.date_range.filter_map do |date|
             if promotion.date_eligible?(date)
-              line_item_per_unit = Struct.new(:amount, :currency, :quantity)
-                                         .new(line_item.amount_per_date_unit, line_item.currency, line_item.quantity)
+              line_item_per_unit = Struct.new(:amount, :currency, :quantity, :price, :cap)
+                                         .new(line_item.amount_per_date_unit, line_item.currency,
+                                              line_item.quantity, line_item.price, calculator&.cap
```
theachoem, line 41: If cap is already in calculator, I think no need to put cap to line item struct object here.

db/migrate/20231117032031_add_cap_to_spree_calculator.rb
```diff
@@ -0,0 +1,5 @@
+class AddCapToSpreeCalculator < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_calculators, :cap, :float, if_not_exists: true
+  end
```
theachoem, line 4: Instead of putting cap on both promotion and calculator, what do you think of putting just one? but in promotion action instead. Some actions may not support cap, so we can do that in action instead.

app/models/spree_cm_commissioner/promotion/actions/create_date_specific_item_adjustments.rb
theachoem: image

app/models/spree_cm_commissioner/calculator/percent_on_line_item_decorator.rb
theachoem: I still think putting caculation in promotion action is more central point. We can put both cap db field and do all caculation there in single place. We only need to decorate 3 class.

- Promotion::Actions::CreateAdjustment
- Promotion::Actions::CreateItemAdjustments
- Promotion::Actions::CreateDateSpecificItemAdjustments

db/migrate/20231117032031_add_cap_to_spree_calculator.rb
kimsrung: let's try

kimsrung:
bong @channainfo , after refactoring code I think we are no longer need cap on promotion. As of now, I still keep my old code but I would like to remove that after discussion.

Also these classes:
- Promotion::Actions::CreateItemAdjustments
- Promotion::Actions::CreateDateSpecificItemAdjustments
are using the same calculator Spree::Calculator::PercentOnLineItem, so override compute function here would affect all.

as of ItemTotal adjustment, we shall modify in this Spree::Calculator::FlatPercentItemTotal

@theachoem, @channainfo we can try as Thea mentioned to modify in these 3 classes:
- Promotion::Actions::CreateAdjustment
- Promotion::Actions::CreateItemAdjustments
- Promotion::Actions::CreateDateSpecificItemAdjustments
but this shall affected in other calculators. 
For example, Promotion::Actions::CreateAdjustment could have many calculators such as flat_rate, flexi_rate, shipping, tiered_percent.... If we override method calculate in this class, I'm not sure if it would be the right place.

theachoem (approved):
Look good bong, i think adding in caculator make more sense if we want to cap each day.

PR #833, opened 2023-11-20
close #832 save queue id to sms log

### LINKED ISSUE #832: Save queue ID to sms log

theachoem:


db/migrate/20231120105233_add_queue_id_to_sms_log.rb
```diff
@@ -0,0 +1,5 @@
+class AddQueueIdToSmsLog < ActiveRecord::Migration[7.0]
+  def change
+    add_column :cm_sms_logs, :queue_id, :string, if_not_exists: true
```
channainfo, line 3: @theachoem let's call this column **external_ref** instead set the length to 36 chars (the uuid length) and add an index to this column

PR #849, opened 2023-11-23
close #848 remove penalty from order summary

### LINKED ISSUE #848: Remove penalty from order summary

theachoem:
Replace penalty label with "adjustment" instead!

image


channainfo:
@theachoem if it is removed, we will see inconsistent value between the subtotal and the total. I think let's keep it there for now and rename it to "adjustment" instead.

theachoem:
@channainfo Done bong!

PR #853, opened 2023-11-23
close #852 send image in notification payload

### LINKED ISSUE #852: Send image in notification payload

panhachom:
#### Documentation 
https://firebase.google.com/docs/cloud-messaging/android/send-image
#### Demo 
notification (1)

### DIFF
--- a/app/models/spree_cm_commissioner/customer_notification.rb
+++ b/app/models/spree_cm_commissioner/customer_notification.rb
@@ -10,8 +10,14 @@ module SpreeCmCommissioner
     validates :payload, presence: true
     validates :notification_type, presence: true
 
+    def default_notification_image_url
+      Spree::Store.default.default_notification_image&.styles&.first&.[](:url)
+    end
+
     def push_notification_image_url
-      image_attrs = feature_image.style(:mini)
+      return default_notification_image_url if feature_image.nil?
+
+      image_attrs = feature_image.styles.first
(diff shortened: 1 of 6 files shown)


app/models/spree_cm_commissioner/customer_notification.rb
```diff
@@ -11,7 +11,9 @@ class CustomerNotification < SpreeCmCommissioner::Base
     validates :notification_type, presence: true
 
     def push_notification_image_url
-      image_attrs = feature_image.style(:mini)
+      return if feature_image.nil?
```
channainfo, line 14: use a default image with image url helper if it does not exist

app/notifications/noticed_fcm_base.rb
```diff
@@ -24,15 +24,47 @@ def fcm_device_tokens(recipient)
...
+      },
+
+      # send notication with image
+      android: android_settings,
+      apns: apns_settings
+    }
+  end
+
+  def image_url
+    nil
```
channainfo, line 43: set a default image

channainfo (changes requested):

app/notifications/noticed_fcm_base.rb
panhachom: @channainfo but bong some notification type do not require image like order_complete_notification or should  all notification require image ?

PR #886, opened 2023-12-09
close #865 update user confirm_at to check unconfirmed user

### LINKED ISSUE #865: Unconfirmed users
Once confirming the confirmation code we should update the confirmed_at to make sure we know how many users are not confirmed.

vaneath:


app/interactors/spree_cm_commissioner/user_pin_code_authenticator.rb
```diff
@@ -40,6 +41,10 @@ def register_user!
       end
     end
 
+    def update_user_pin_code_confirmed_at!
+      Spree::User.where(id: context.user.id).update(confirmed_at: Time.zone.now)
```
theachoem, line 45: In devise, we can just use skip_confirmation! Call it before creating.

`context.user.skip_confirmation!`

Learn more how devise confirmation work:
https://github.com/heartcombo/devise/blob/main/lib/devise/models/confirmable.rb

vaneath: Solve it with manually set confirmed_at with Time.zone.now

app/interactors/spree_cm_commissioner/user_pin_code_authenticator.rb
```diff
@@ -32,6 +32,7 @@ def register_user!
       create_options = context.to_h.slice(*fields)
 
       user = Spree::User.new(create_options)
+      user[:confirmed_at] = Time.zone.now
```
channainfo, line 35: use **user. confirmed_at** instead. Add a spec to make sure this field is assigned correctly

channainfo (changes requested):

PR #889, opened 2023-12-12
close #888 read all notification api

### LINKED ISSUE #888: read all notification api

panhachom:
image

https://www.figma.com/file/lozZfapNXqz2IQQkz7kQiY/Accomodation-App---Dev?type=design&node-id=1755-9948&mode=design&t=NXO4Glv1UyRBjsPf-0

### DIFF
--- a/app/controllers/spree/api/v2/storefront/notifications_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/notifications_controller.rb
@@ -30,7 +30,19 @@ module Spree
           end
 
           def mark_as_read
-            SpreeCmCommissioner::NotificationReader.call(id: params[:id])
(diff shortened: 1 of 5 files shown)


app/models/spree_cm_commissioner/customer_notification.rb
```diff
@@ -10,7 +10,7 @@ class CustomerNotification < SpreeCmCommissioner::Base
     validates :payload, presence: true
     validates :notification_type, presence: true
 
-    enum notification_type: { :promotion => 0, :announcement => 1 }
+    enum notification_type: { :promotion_notification => 0, :announcement_notification => 1 }
```
channainfo, line 13: let's keep the old naming. I think the promotion_notification is a bit redundant.

channainfo (changes requested):

PR #903, opened 2023-12-15
close #902 add default state to vendor serializer

### LINKED ISSUE #902: Add default state to vendor serializer

theachoem:


app/serializers/spree/v2/storefront/accommodation_serializer.rb
```diff
@@ -7,8 +7,6 @@ class AccommodationSerializer < VendorSerializer
         # because its model still Spree::Vendor
         set_type :vendor
 
-        has_one :default_state, serializer: :state
```
channainfo, line 10: have you checked if we use this anymore to make sure it is backward compatible.

channainfo (changes requested):

app/serializers/spree/v2/storefront/accommodation_serializer.rb
theachoem: In current develop branch in app, we required it bong.

PR #905, opened 2023-12-16
close #904 add include kyc to spree product

### LINKED ISSUE #904: add include_kyc boolean to spree_product

panhachom:


db/migrate/20231215083209_add_include_kyc_to_spree_product.rb
```diff
@@ -0,0 +1,5 @@
+class AddIncludeKycToSpreeProduct < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_products , :include_kyc, :boolean, default: false, if_not_exists: true
```
channainfo, line 3: let's avoid calling fields with the include_ prefix. rename it to  guest_kyc instead.

channainfo (changes requested):

PR #923, opened 2023-12-21
close #922 add kyc field to spree product

### LINKED ISSUE #922: Add kyc_field to spree product

panhachom:
https://github.com/channainfo/commissioner/assets/92453740/728830b6-03f0-4e5a-9ec7-9380f0cad73e


db/migrate/20231220071350_add_kyc_field_to_spree_products.rb
```diff
@@ -0,0 +1,5 @@
+class AddKycFieldToSpreeProducts < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_products, :kyc_field, :integer, if_not_exists: true
```
channainfo, line 3: add a default value to 0. why don't just use the existing kyc field by changing the column from boolean to integer

app/models/concerns/spree_cm_commissioner/kyc_field_bitwise.rb
```diff
@@ -0,0 +1,23 @@
...
+    BIT_FIELDS = {
+      customer_name: 0b00001,
+      customer_gender: 0b00010,
+      customer_dob: 0b00100,
+      customer_occupation: 0b01000,
+      customer_id_card: 0b10000
+    }.freeze
+
+    def include_kyc?
+      kyc_field.present?
```
channainfo, line 14: let's use need_kyc? instead. I think we have many need_  in our code base.
present? for zero is true so make sure it is present and not zero. Probably
!kyc_field.zero?

channainfo (changes requested):

app/models/concerns/spree_cm_commissioner/kyc_field_bitwise.rb
panhachom: done bong

channainfo:
@panhachom failed!!!

panhachom:
> @panhachom failed!!!
resolved bong

app/interactors/spree_cm_commissioner/kyc_calculator.rb
```diff
@@ -0,0 +1,16 @@
+module SpreeCmCommissioner
+  class KycCalculator < BaseInteractor
+    delegate :params, to: :context
+
+    def call
```
channainfo, line 5: This method has no interactions and failure status so simple so let's just put in a **module** and include it instead. Let's add a spec for this as well.

app/models/concerns/spree_cm_commissioner/kyc_bitwise.rb
```diff
@@ -0,0 +1,23 @@
...
+      customer_gender: 0b00010,
+      customer_dob: 0b00100,
+      customer_occupation: 0b01000,
+      customer_id_card: 0b10000
+    }.freeze
+
+    def need_kyc?
+      !kyc.zero?
+    end
+
```
channainfo, line 16: Probably merge the kyc calculator with this one and call it kycable.rb

config/routes.rb
```diff
@@ -68,6 +68,11 @@
     end
 
     resources :products do
+      resources :kyc, only: %i[index update] do
```
channainfo, line 71: here KYC is not a resource/resources, I think it would be better to call it get :edit_kyc and put :update_kyc under the products controller

channainfo (changes requested):

app/models/concerns/spree_cm_commissioner/kyc_bitwise.rb
panhachom: Since we have set kyc default value to 0 we don't need this method anymore

PR #935, opened 2023-12-27
close #931 guest id card api

### LINKED ISSUE #931: Guest Id card api

panhachom:
image

### DIFF
--- a/app/controllers/spree/api/v2/storefront/id_cards_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/id_cards_controller.rb
@@ -1,0 +1,63 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class IdCardsController < ::Spree::Api::V2::ResourceController
+          def create
+            context = SpreeCmCommissioner::GuestIdCardManager.call(
+              card_type: id_card_params[:card_type],
+              front_image_url: id_card_params[:front_image_url],
+              back_image_url: id_card_params[:back_image_url],
+              guest_id: id_card_params[:guest_id]
+            )
+
+            if context.success?
+              render_serialized_payload { serialize_resource(context.result) }
+            else
+              render_error_payload(context.message)
+            end
+          end
+
+          def update
+            context = SpreeCmCommissioner::GuestIdCardManager.call(
+              card_type: id_card_params[:card_type],
+              front_image_url: id_card_params[:front_image_url],
+              back_image_url: id_card_params[:back_image_url],
+              guest_id: id_card_params[:guest_id]
+            )
+
+            if context.success?
+              render_serialized_payload { serialize_resource(context.result) }
+            else
+              render_error_payload(context.message)
+            end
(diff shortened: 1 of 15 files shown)


app/helpers/spree_cm_commissioner/admin/s3_helper.rb
```diff
@@ -0,0 +1,22 @@
+module SpreeCmCommissioner
+  module Admin
+    module S3Helper
```
channainfo, line 3: Don't use the Helper suffix. Helper are meant to be used in the view. We can move this to a libs/s3_url_generator.rb and create a self.presigned_url method instead.

app/controllers/spree/api/v2/storefront/id_cards_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class IdCardsController < ::Spree::Api::V2::ResourceController
+          before_action :load_guest
+
+          def update
+            id_card = @guest.id_card || SpreeCmCommissioner::IdCard.new(guest: @guest)
```
channainfo, line 9: separate the create and update. If the guest does not have id_card then it should be sent to create otherwise it will be the update one.

channainfo (changes requested):

app/controllers/spree/api/v2/storefront/id_cards_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+          private
+
+          def id_card_params
+            params.permit(:card_type, :front_image_url, :back_image_url).merge(guest: @guest)
+          end
+
+          def load_resource
+            @guest ||= SpreeCmCommissioner::Guest.find(params[:guest_id])
+            @id_card ||= @guest.id_card
```
channainfo, line 38: remove this, it is confusing because it is only being used in the update.

app/controllers/spree/api/v2/storefront/id_cards_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  module Api
+    module V2
+      module Storefront
+        class IdCardsController < ::Spree::Api::V2::ResourceController
+          before_action :load_resource
+
+          def create
+            id_card = SpreeCmCommissioner::IdCard.new(guest: @guest)
+
+            context = SpreeCmCommissioner::IdCardModifier.call(options: id_card_params, id_card: id_card)
```
channainfo, line 11: I feel this method is a bit confusing because when using the POST create it does not have the id_card.

Let's rename the class to GuestIdCard and then instead of receiving the **option** and the **id_card**,  use **option** only. In the class you need to check you save the cards if the front_image_url and back_image_url are present. You should also have an api for client to remove the id card by passing a parameter to let the server know which card to remove.

channainfo (changes requested):

PR #978, opened 2024-01-08
Close #975 Display Vehicle/Vehicle-code in edit page

### LINKED ISSUE #975: In editing page display Vehicle/vehicle_code instead of just vehicle_code (ex: Vehicle/ARB1)

Regene27:
image
image

### DIFF
--- a/app/models/spree_cm_commissioner/vehicle.rb
+++ b/app/models/spree_cm_commissioner/vehicle.rb
@@ -2,6 +2,8 @@ require_dependency 'spree_cm_commissioner'
 
 module SpreeCmCommissioner
   class Vehicle < SpreeCmCommissioner::Base
+    has_one :primary_photo, -> { order(position: :asc) }, class_name: 'SpreeCmCommissioner::VehiclePhoto', as: :viewable, dependent: :destroy
+
     belongs_to :vehicle_type, class_name: 'SpreeCmCommissioner::VehicleType', dependent: :destroy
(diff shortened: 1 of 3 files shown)


channainfo:
@Regene27 resolve conflict

app/models/spree_cm_commissioner/vehicle.rb
```diff
@@ -13,5 +13,9 @@ class Vehicle < SpreeCmCommissioner::Base
 
     self.whitelisted_ransackable_attributes = %w[license_plate code]
     self.whitelisted_ransackable_associations = %w[vehicle_type]
+
+    def primary_photo
+      self.vehicle_photo.find_by(position: 1)
```
channainfo, line 18: don't do this. It will cause n+1 query. Use the custom relationship instead.

channainfo (changes requested):

app/models/spree_cm_commissioner/vehicle.rb
```diff
@@ -2,6 +2,8 @@
 
 module SpreeCmCommissioner
   class Vehicle < SpreeCmCommissioner::Base
+    has_one :primary_photo, -> { order(:position) }, class_name: 'SpreeCmCommissioner::VehiclePhoto', as: :viewable, dependent: :destroy
```
channainfo, line 5: explicitly state ASC or DESC in the order.

channainfo (changes requested):

PR #993, opened 2024-01-09
Close #979 Add Primary Photo to Vehicle

### LINKED ISSUE #979: should be able to select an image as a primary photo of a vehicle

Regene27:
image

### DIFF
--- a/app/views/spree/transit/vehicles/index.html.erb
+++ b/app/views/spree/transit/vehicles/index.html.erb
@@ -13,6 +13,7 @@
     <table class="table" id="listing_vehicles" data-hook>
       <thead class="text-muted">
         <tr data-hook="transit_vehicle_index_headers">
+          <th><%= Spree.t('transit.primary_photo')%></th>
           <th>
             <%= sort_link @search, :code, Spree.t('transit.vehicle.code') %>
           </th>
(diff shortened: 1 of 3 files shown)


app/views/spree/transit/vehicles/index.html.erb
```diff
@@ -24,6 +25,12 @@
       <tbody>
         <% @vehicles.each do |vehicle| %>
           <tr data-hook="transit_vehicle_index_rows">
+            <td>
+                <div data-hook="preview-image">
+                  <%= link_to image_tag(main_app.url_for((vehicle.vehicle_photo.first.attachment.variant(resize_to_fill: [60, 60]))), :class => "rounded border mb-4"),
+                              main_app.rails_blob_url(vehicle.vehicle_photo.first.attachment) %>
```
tykeaboyloy, line 31: I think it's better to define a method for primary photo inside the model instead of view
```rb
#vehicle.rb
    def primary_photo
      self.vehicle_photo.find_by(position: 1)
    end
```
then call it in the view
```rb
#vehicles/index.html.erb
 <%= image_tag main_app.url_for(vehicle.primary_photo.attachment) if vehicle.primary_photo %>
```

PR #1048, opened 2024-01-25
close #1047 validate quantity if it's allowed to decrease

### LINKED ISSUE #1047: Validate quantity if it is allowed to decrease in line item
PR #1048

vaneath:
In mobile, if user try to decrease while having guest, it should raise error and user must delete some guest first.


theachoem:
Close as no longer needed.

PR #1072, opened 2024-02-01
Close #1061 Add Account Name and Account Number to Spree Vendors

### LINKED ISSUE #1061: Samram Project - Reporting and Mark to Paid from Report
https://drive.google.com/drive/u/1/folders/1zye9MW1bnjsZ8H5fV4ZOopFdo45joxvx

Please follow the four report sampling in the Google Sheet, most of it already exists we only need to add a bit more field into the report.

The new feature they wanted to be added in is a 

Screenshot 2024-01-30 185402

Put the bank detail below the QR inside the invoice

Screenshot 2024-01-30 185315

Screenshot 2024-01-30 185119

They want the Marked to Paid button in the Balance Due Report too. 

Lastly, the main currency type must have in KHR (Khmer Riel).

Regene27:
image
image

### DIFF
--- a/app/models/spree_cm_commissioner/vendor_decorator.rb
+++ b/app/models/spree_cm_commissioner/vendor_decorator.rb
@@ -10,6 +10,8 @@ module SpreeCmCommissioner
 
       base.attr_accessor :service_availabilities
 
+      base.store :preferences, accessors: %i[account_name account_number]
+
       base.before_save :generate_code, if: :code.nil?
 
       base.has_many :photos, -> { order(:position) }, as: :viewable, dependent: :destroy, class_name: 'SpreeCmCommissioner::VendorPhoto'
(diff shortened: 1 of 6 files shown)


app/controllers/spree/billing/vendors_controller.rb
```diff
@@ -26,6 +26,15 @@ def update_positions
         end
       end
 
+      def remove_payment_qrcode
```
channainfo, line 29: create another controller called payment_qrcode and use the destroy instead.

```
  resources :vendors do
     resource :payment_qrcodes, only: %i[destroy]
  end
```

db/migrate/20240201072700_add_account_name_and_account_number_to_spree_vendors.rb
```diff
@@ -0,0 +1,6 @@
+class AddAccountNameAndAccountNumberToSpreeVendors < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_vendors, :account_name, :string, if_not_exists: true
```
channainfo, line 3: let's not create these two columns. Use the preference to store them instead.

channainfo (changes requested):

app/models/spree_cm_commissioner/vendor_decorator.rb
```diff
@@ -148,6 +152,11 @@ def selected_option_value_vendors_ids
     def generate_code
       self.code = (code.presence || name[0, 3].upcase)
     end
+
+    def validate_account_name_and_number_presence
```
channainfo, line 156: make sure this validation is needed only for the billing vendor.

channainfo (changes requested):

PR #1085, opened 2024-02-05
WIP #1084 add update button to update the updated_at field for homepage section

### LINKED ISSUE #1084: Add update button to update the updated_at field for homepage section

vaneath:


app/controllers/spree/admin/homepage_section_controller.rb
```diff
@@ -13,12 +13,20 @@ def collection_url(options = {})
...
 
+      # ovverride
       def edit
         homepage_section_id = params[:id]
         @homepage_section = SpreeCmCommissioner::HomepageSection.find(homepage_section_id)
         @homepage_section_relatables = SpreeCmCommissioner::HomepageSectionRelatable.where(homepage_section: @homepage_section)
       end
 
+      # This method update the updated_at field of all homepage sections
+      def update_the_updated_at
```
channainfo, line 24: why we need this?

channainfo (changes requested):

app/controllers/spree/admin/homepage_section_controller.rb
vaneath: Bong thea said sections are cached once it's loaded, and only clear after the updated_at field is updated. This button will update the updated_at to Time.zone.now so operator can manually update all sections to latest setup, without any glitch of cache.

channainfo: Let's explicitly clear the cache instead. Create an item under the Settings called "Caches" which lists all the cache along with actions to clear the cache. In the meantime, we just need to list down the most important caches to be cleared.

vaneath:
close as not currently needed.

PR #1086, opened 2024-02-05
Close #1073 Add Invoice Capture Button to Balance Due Report page

### LINKED ISSUE #1073: Add Invoice Capture button to Balance Due Report page

Regene27:
image


app/controllers/spree/billing/payments_controller.rb
```diff
@@ -40,6 +40,22 @@ def show
         @payment = @object
       end
 
+      def fire
+        return unless @payment.payment_source
```
channainfo, line 44: if the condition is met, then this will be silently failed. You should raise and handle an exception in the **rescue_from** block

app/controllers/spree/billing/payments_controller.rb
```diff
@@ -40,6 +40,22 @@ def show
         @payment = @object
       end
 
+      def fire
```
channainfo, line 43: why not using the **update** action instead?. Remember to avoid using a weird **action** name

app/controllers/spree/billing/payments_controller.rb
```diff
@@ -40,6 +40,22 @@ def show
...
+        return unless @payment.payment_source
+
+        # Because we have a transition method also called void, we do this to avoid conflicts.
+        params[:e] = 'void_transaction' if params[:e] == 'void'
+        if @payment.send("#{params[:e]}!")
+          flash[:success] = Spree.t(:payment_updated)
+        else
+          flash[:error] = Spree.t(:cannot_perform_operation)
+        end
+      rescue Spree::Core::GatewayError => e
```
channainfo, line 53: Exception should be handled with the **rescue_from** block

channainfo (changes requested):

app/controllers/spree/billing/payments_controller.rb
```diff
@@ -40,12 +43,46 @@ def show
         @payment = @object
       end
 
+      def update
+        check_payment_source
+
+        # Because we have a transition method also called void, we do this to avoid conflicts.
+        params[:e] = 'void_transaction' if params[:e] == 'void'
```
channainfo, line 50: make sure to check the value with the whitelisted value if the params pass from the client to run the code dynamically using the send method.

channainfo (changes requested):

app/controllers/spree/billing/payments_controller.rb
```diff
@@ -40,12 +43,52 @@ def show
...
+        if supported_events.include?(params[:e])
+          if @payment.send("#{params[:e]}!")
+            flash[:success] = Spree.t(:payment_updated)
+          else
+            flash[:error] = Spree.t(:cannot_perform_operation)
+            raise Spree::Core::GatewayError
+          end
+        else
+          flash[:error] = Spree.t(:unsupported_event)
+          raise Spree::Core::GatewayError
```
channainfo, line 62: Don't raise an exception in the controller action and avoid the nested condition. You can use before_action to do that.

channainfo (changes requested):

PR #1095, opened 2024-02-07
Close #1087 Modify Template for Billing Report

### LINKED ISSUE #1087: Billing Report modify template

Regene27:
image
image
image
image

### DIFF
--- a/app/controllers/spree/billing/invoice_controller.rb
+++ b/app/controllers/spree/billing/invoice_controller.rb
@@ -12,6 +12,10 @@ module Spree
         redirect_back fallback_location: admin_path
       end
 
+      def print
+        order.update(approved_at: Time.zone.now)
+      end
+
       # @overrided
       def order
         @order = Spree::Order.find_by!(number: params[:order_id])
--- a/app/controllers/spree/billing/report_controller.rb
+++ b/app/controllers/spree/billing/report_controller.rb
@@ -15,11 +15,23 @@ module Spree
       end
 
       def paid
+        if params[:q].present?
+          params[:q][:updated_at_lteq] = params[:q][:updated_at_lteq].to_date.end_of_day if params[:q][:updated_at_lteq].present?
+        else
+          params[:q] = {}
+        end
+
         @search = Spree::Order.subscription.ransack(params[:q])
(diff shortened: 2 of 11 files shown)


app/models/spree_cm_commissioner/order_decorator.rb
```diff
@@ -138,6 +138,12 @@ def qr_data
       "#{number}-#{token}"
     end
 
+    def payment_date
```
channainfo, line 141: if you consider updated_at is the payment_date will mislead. Updated at receives new a new value for every update done to the record.

app/controllers/spree/billing/invoice_controller.rb
```diff
@@ -12,6 +12,10 @@ def create
         redirect_back fallback_location: admin_path
       end
```
channainfo, line 14: When you have custom actions always add the HTTP verb and the routes right on top of the action e.g
```
# PUT /your_route_path
def print
...

```


```

**Review by channainfo — CHANGES_REQUESTED** [review]

File: app/controllers/spree/billing/invoice_controller.rb [hunk 3]
```diff
@@ -12,6 +12,10 @@ def create
...
 
+      def print
```
channainfo, line 15: This is misleading. The action is print but you update the approved_at

app/controllers/spree/billing/report_controller.rb
```diff
@@ -15,11 +15,23 @@ def show
       end
 
       def paid
+        if params[:q].present?
```
channainfo, line 18: Move this logic to a service. This one is very important for business. It need to be tested well with a spec

app/controllers/spree/billing/report_controller.rb
```diff
@@ -15,11 +15,23 @@ def show
...
+        else
+          params[:q] = {}
+        end
+
         @search = Spree::Order.subscription.ransack(params[:q])
         @orders = @search.result.where(payment_state: 'paid').page(page).per(per_page)
       end
 
       def balance_due
+        if params[:q].present?
```
channainfo, line 29: and reuse the service above.

channainfo (changes requested):

PR #1098, opened 2024-02-07
close #1097 add section_type column to homepage_section table

### LINKED ISSUE #1097: Add segment column to homepage_section table
PR #1110

vaneath:
Issue #1097 

Screenshot from 2024-02-07 15-52-41

### DIFF
--- a/app/models/spree_cm_commissioner/homepage_section.rb
+++ b/app/models/spree_cm_commissioner/homepage_section.rb
@@ -2,6 +2,11 @@ module SpreeCmCommissioner
   class HomepageSection < Base
     acts_as_list column: :position
 
+    enum section_type: {
+      general: 0,
+      ticket: 1
+    }
+
     attribute :title
     attribute :description
     attribute :active
--- a/app/views/spree/admin/homepage_section/_form.html.erb
+++ b/app/views/spree/admin/homepage_section/_form.html.erb
@@ -8,6 +8,10 @@
             <%= f.label :title %>
             <%= f.text_field :title, class: 'form-control' %>
           <% end %>
+          <%= f.field_container :section_type do %>
+            <%= f.label :section_type %>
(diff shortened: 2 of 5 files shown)


app/models/spree_cm_commissioner/homepage_section.rb
```diff
@@ -2,6 +2,11 @@ module SpreeCmCommissioner
   class HomepageSection < Base
     acts_as_list column: :position
 
+    enum section_type: {
```
channainfo, line 5: Let's call it "segment" instead store it as an integer and use it as a bitwise instead of the enum.

app/views/spree/admin/homepage_section/_form.html.erb
```diff
@@ -8,6 +8,10 @@
             <%= f.label :title %>
             <%= f.text_field :title, class: 'form-control' %>
           <% end %>
+          <%= f.field_container :section_type do %>
+            <%= f.label :section_type %>
+            <%= f.select :section_type, SpreeCmCommissioner::HomepageSection.section_types.keys, {}, { class: 'fullwidth select2' } %>
```
channainfo, line 13: use a checkbox and allow user to enter different at least of the the values

channainfo (changes requested):

app/views/spree/admin/homepage_section/_form.html.erb
vaneath: noted b, my current flow using select type so that each section knows which homepage it belongs to, but if using bitwise and checkbox can be more flexible since each section may belong to more than 1 homepage. I will check that!

channainfo: The **segment** and **section** are interchangeable. The term "type" tends to have a string value instead of a numeric value, thus we drop the **type** out. Segment tends to have more tendency toward a numeric value.

vaneath:
Moved to #1110

PR #1108, opened 2024-02-13
wip #1107 add started_at to spree_product

### LINKED ISSUE #1107: Add started_at to spree_product

panhachom:
image

### Usage 
- This field is needed to keep track of time of upcoming product so that we can perform schedule push notification  to alert user


channainfo:
@panhachom why do we need this?

panhachom:
> @panhachom why do we need this? let's add the purpose of having this field in the description as well.

noted bong

PR #1110, opened 2024-02-14
close #1097 add segment bitwise to homepage_sections table

### LINKED ISSUE #1097: Add segment column to homepage_section table

vaneath:
[Screencast from 2024-02-15 15-18-25.webm](https://github.com/channainfo/commissioner/assets/109834020/bead12fc-f132-4918-839a-34073240f232)

### DIFF
--- a/app/helpers/spree_cm_commissioner/admin/homepage_segment_helper.rb
+++ b/app/helpers/spree_cm_commissioner/admin/homepage_segment_helper.rb
@@ -1,0 +1,30 @@
+module SpreeCmCommissioner
+  module Admin
+    module HomepageSegmentHelper
+      def self.badge_class_for_segment(segment_name)
+        case segment_name.to_sym
+        when :general
+          'badge badge-primary text-uppercase'
+        when :ticket
+          'badge badge-secondary text-uppercase'
+        when :tour
+          'badge badge-info text-uppercase'
+        when :accommodation
+          'badge badge-warning text-uppercase'
+        else
+          'badge'
+        end
+      end
+
+      def calculate_segment_value(params)
+        segment_params = params.slice(*SpreeCmCommissioner::HomepageSectionBitwise::BIT_SEGMENT.keys)
+
+        return nil unless segment_params.values.any?
+
+        segment_params.values.each_with_index.sum do |value, index|
+          value.to_i * (2**index)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


channainfo:
@vaneath After a new section is created, redirect to edit instead because it makes more sense to add Relatable items. The new section is useless without any relatables. Also, add a counter_cache to the section relatable. Display the number of section relatable count in the table. Make sure we only return section where it section relatable count is great than 1 ( using the counter_cache)

vaneath:
> @vaneath After a new section is created, redirect to edit instead because it makes more sense to add Relatable items. The new section is useless without any relatables. Also, add a counter_cache to the section relatable. Display the number of section relatable count in the table. Make sure we only return section where it section relatable count is great than 1 ( using the counter_cache)

noted bong, I will check on it

app/models/concerns/spree_cm_commissioner/homepage_section_bitwise.rb
```diff
@@ -0,0 +1,26 @@
+module SpreeCmCommissioner
+  module HomepageSectionBitwise
+    extend ActiveSupport::Concern
+
+    BIT_SEGMENT = {
+      general: 0b00001,
```
theachoem, line 6: Let add tour & accommodation as well. `general, ticket, tour, accommodation`

app/helpers/spree_cm_commissioner/admin/homepage_segment_helper.rb
```diff
@@ -0,0 +1,14 @@
+module SpreeCmCommissioner
+  module Admin
+    module HomepageSegmentHelper
+      def calculate_segment_value(params)
+        segment_params = params.slice(:general, :ticket)
```
theachoem, line 5: params.slice(*SpreeCmCommissioner::HomepageSectionBitwise:: BIT_SEGMENT.keys)

theachoem (changes requested):

app/models/spree_cm_commissioner/homepage_section.rb
```diff
@@ -1,14 +1,24 @@
...
-    attribute :description
-    attribute :active
-    attribute :position
+    attribute :segment, :integer, default: 0
 
     has_many :homepage_section_relatables, -> { active }, inverse_of: :homepage_section, dependent: :destroy
 
     scope :active, -> { where(active: true).order(:position) }
+
+    def self.badge_class_for_segment(segment_name)
```
theachoem, line 13: Move this to controller helper

theachoem (changes requested):

PR #1126, opened 2024-02-16
close #1118 segment homepage image background for specific page

### LINKED ISSUE #1118: Add new api route for homepage background

vaneath:
- Previously we can only upload image for home screen. 
- In this PR, we added new field call `segment` to homepage_backgrounds, and allow frontend to fetch that image for a specific screen.

[Screencast from 2024-02-16 10-48-24.webm](https://github.com/channainfo/commissioner/assets/109834020/7c64cafd-2c2e-4003-97d4-3c7affeb6210)

### DIFF
--- a/app/controllers/spree/admin/homepage_section_controller.rb
+++ b/app/controllers/spree/admin/homepage_section_controller.rb
@@ -1,7 +1,7 @@
 module Spree
   module Admin
     class HomepageSectionController < Spree::Admin::ResourceController
-      include SpreeCmCommissioner::Admin::HomepageSegmentHelper
+      helper SpreeCmCommissioner::Admin::HomepageSegmentHelper
 
       def model_class
         SpreeCmCommissioner::HomepageSection
@@ -46,7 +46,7 @@ module Spree
 
       # overrided
       def permitted_resource_params
-        segment_value = calculate_segment_value(params[:spree_cm_commissioner_homepage_section])
+        segment_value = helpers.calculate_segment_value(params[:spree_cm_commissioner_homepage_section])
 
         params.require(:spree_cm_commissioner_homepage_section).permit(:title, :description, :active).merge(segment: segment_value)
       end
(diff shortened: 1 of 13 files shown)


channainfo:
Resolve conflicts

app/controllers/spree/admin/homepage_section_controller.rb
```diff
@@ -2,6 +2,7 @@ module Spree
   module Admin
     class HomepageSectionController < Spree::Admin::ResourceController
       include SpreeCmCommissioner::Admin::HomepageSegmentHelper
+      helper_method :badge_class_for_segment
```
channainfo, line 5: remove this ```helper_method```. If you include the helper module, all its methods will be available to the controller and the view.

channainfo (changes requested):

app/controllers/spree/admin/homepage_section_controller.rb
vaneath: Helper module not working with include, now I change to helper instead of instead since it's a helper module. It should work not bong!

vaneath: I use badge_class_for_segment for view and now I move calculate_segment_value to controller now it should work fine bong.

vaneath: what is the best practice bong since:
include SpreeCmCommissioner::Admin::HomepageSegmentHelper will only available in controller scope. In order to use it in view I have to: helper SpreeCmCommissioner::Admin::HomepageSegmentHelper instead of include or define it like this:       helper_method :badge_class_for_segment.

vaneath: I read from this bong: https://stackoverflow.com/questions/5130150/rails-how-to-use-a-helper-inside-a-controller

channainfo: If you name your helper after your controller name, the methods will be available on both controller and the view. For the  Admin::HomepageSectionController the helper would be Admin::HomepageSectionHelper.

PR #1132, opened 2024-02-16
close #1127 fixed billing locales to only contain en|km

### LINKED ISSUE #1127: Billing routes include many unnecessary locales
image

tykealy:


config/routes.rb
```diff
@@ -163,7 +163,7 @@
     get '/resource_not_found', to: 'errors#resource_not_found'
   end
 
-  scope '(:locale)', locale: /#{I18n.available_locales.join('|')}/ do
```
channainfo, line 166: why replacing this with a hard coded one.

channainfo (changes requested):

PR #1175, opened 2024-02-29
close #1172 add hidden boolean to spree option type

### LINKED ISSUE #1172: add hidden boolean to spree_option_type

panhachom:
-  Use Hidden checkbox to filter out hidden option type such Reminder or Started At 
image


theachoem (changes requested):
When return variants to app, should exclude hidden as well. So app doesn't show the hidden option type & value.

panhachom:
> When return variants to app, should exclude hidden as well. So app doesn't show the hidden option type & value.

i have already created visible_option_values method which exclude the hidden one 

> When return variants to app, should exclude hidden as well. So app doesn't show the hidden option type & value.

@theachoem  done bong

channainfo:
@panhachom check the test is broken

Issue #1151, opened 2024-02-21
Pricing Rate & Availability


theachoem:

theachoem:
![image](https://github.com/channainfo/commissioner/assets/29684683/6c66a1e5-6079-405e-be20-638262373d78)

theachoem:
![image](https://github.com/channainfo/commissioner/assets/29684683/53735ebe-2b64-48e3-88e6-7442600117ce)

theachoem:
![image](https://github.com/channainfo/commissioner/assets/29684683/8a5e99c1-6fa9-4590-a5e3-ba6236fed969)

theachoem:
## Channel Manager 1 : Siteminder
https://help.siteminder.com/s/article/Set-up-and-manage-availability?language=en_US

![image](https://github.com/channainfo/commissioner/assets/29684683/e3a6bc3a-a5c8-4a3e-b19f-c5d57deaf40c)

theachoem:
- vendor has many products
- product has many variants
- variant has many prices

1. Should price have rates?
Yes, price should have rates. Anyways, a variant can have many prices base on currency, so it is better to move rates to variant instead.

2. Should variants have rates?
The rate in variants is good as it can be reused for all currencies.

3. Should vendors have rates?
The rate in products can be used across all variants. This might be useful as free-cancellation

5. Should product have rates?

theachoem:
Rate will be a polymeric and it currently should only work with variants. 

### Rate
- Location: Rates may vary depending on the location of the property, with prices typically higher in prime locations or tourist destinations.
- Size: Larger rooms or suites may command higher rates compared to standard rooms.
- Market Demographics: Understanding the demographics of the target market can influence pricing decisions. For example, rates might be adjusted to appeal to business travelers versus leisure travelers.
- Level of Competition: Rates may be adjusted relative to competitors' pricing to remain competitive in the market.
- Type of Service Offered: Properties offering additional amenities or services may justify higher rates.
- Day of Week: Rates may fluctuate based on the day of the week, with higher rates on weekends or during peak demand periods.
- Day of Month: Pricing strategies may account for seasonal variations, events, or holidays that affect demand.
- Specific Days: Special events, holidays, or local festivals may warrant adjustments to rates for specific dates.
- Group Booking: Special rates or discounts may apply for group bookings or events.
- These considerations represent the specific factors that inform the determination of rates within the chosen pricing model.

### Pricing Model
[...]

PR #1191, opened 2024-03-05
WIP #1151 add pricing rate model

theachoem:
Brainstorming how rate and model should be applied, will be split into small PR. Brainstorming on issue: #1151


theachoem:
Close to keep old approach as reference.

PR #1196, opened 2024-03-11
close #1195 export guest data to csv

### LINKED ISSUE #1195: Export guest datas to CSV

vaneath:
User can get the csv file by adding extension of **.csv**.  Example: _http://127.0.0.1:4000/events/events-tedxphnompenh-2024/guests.csv_

Raw file in text editor after export:
image

In google sheet:
image

### DIFF
--- a/app/controllers/spree/events/guests_controller.rb
+++ b/app/controllers/spree/events/guests_controller.rb
@@ -11,16 +11,35 @@ module Spree
         SpreeCmCommissioner::Guest
       end
 
+      def index
+        file_name = Rails.root.join('tmp', csv_name)
+
+        respond_with(collection) do |format|
+          format.csv do
+            SpreeCmCommissioner::GenerateGuestsCsv.call(
+              collection: collection,
+              file_name: file_name
+            )
+
+            send_file file_name
+          end
+        end
+      end
+
       def edit
         @guest = SpreeCmCommissioner::Guest.find(params[:id])
         @event = @guest.event
       end
 
+      private
+
+      def csv_name
+        @csv_name ||= "guests-data-#{current_event.name.downcase.gsub(' ', '-')}-#{Time.current.to_i}.csv"
+      end
+
       def permitted_resource_params
         params.required(:spree_cm_commissioner_guest).permit(:entry_type)
       end
-
-      protected
 
       def collection
         scope = model_class.where(event_id: current_event.id)
--- a/spec/interactors/spree_cm_commissioner/generate_guests_csv_spec.rb
+++ b/spec/interactors/spree_cm_commissioner/generate_guests_csv_spec.rb
@@ -1,0 +1,23 @@
+require 'spec_helper'
+require 'csv'
+
+RSpec.describe SpreeCmCommissioner::GenerateGuestsCsv, '.call' do
+  subject(:context) { described_class.call(collection: collection, file_name: file_name) }
+
+  let(:guest1) { create(:guest, first_name: 'John', last_name: 'Doe', dob: Date.new(1990, 1, 1), gender: 1, occupation: create(:taxon, name: 'Engineer'), id_card: create(:id_card, card_type: 0), entry_type: 1) }
+  let(:guest2) { create(:guest, first_name: 'Jane', last_name: 'Smith', dob: Date.new(1985, 5, 5), gender: 2, occupation: create(:taxon, name: 'Doctor'), id_card: create(:id_card, card_type: 1), entry_type: 0) }
+  let(:collection) { SpreeCmCommissioner::Guest.where(id: [guest1.id, guest2.id]) }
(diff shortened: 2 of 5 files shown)


channainfo (changes requested):

app/interactors/spree_cm_commissioner/generate_csv.rb
```diff
@@ -0,0 +1,17 @@
+module SpreeCmCommissioner
+  class GenerateCsv < BaseInteractor
+    delegate :headers, :row_datas, to: :context
+
+    def call
+      csv_data = CSV.generate(headers: true) do |csv|
```
channainfo, line 6: we prefer generate the csv file here. Once it is done, return the file to be sent to the client in the controller.

app/controllers/spree/events/guests_controller.rb
```diff
@@ -11,17 +11,39 @@ def model_class
...
+                [
+                  guest.full_name,
+                  guest.dob&.strftime('%d %b %Y'),
+                  guest.gender&.titleize,
+                  guest.other_occupation&.titleize || guest.occupation&.name,
+                  guest.id_card&.card_type&.titleize,
+                  SpreeCmCommissioner::CheckIn.entry_types.key(guest.entry_type).titleize
+                ]
+              end
+            ).result
```
channainfo, line 29: Don't load this into memory. it will be costly. If there are many records we will face issues. Generate a CSV file in the GenerateCsv instead.

channainfo (changes requested):

app/interactors/spree_cm_commissioner/generate_csv.rb
```diff
@@ -0,0 +1,17 @@
...
+        csv << headers
+
+        row_datas.each do |item|
+          csv << item
+        end
+      end
+
+      context.result = csv_data
```
channainfo, line 14: Open a csv to write data into a file instead of returning a csv data. https://ruby-doc.org/stdlib-2.4.6/libdoc/csv/rdoc/CSV.html#method-c-open

channainfo (changes requested):

app/controllers/spree/events/guests_controller.rb
```diff
@@ -11,17 +11,31 @@ def model_class
         SpreeCmCommissioner::Guest
       end
 
+      def index
+        respond_with(collection) do |format|
+          format.csv do
+            file_name = "#{current_event.name.gsub(' ', '_')}_Guests_Data.csv"
```
channainfo, line 17: construct the absolute file name here and use the lowercase file name and we need a meta-data to know when the file is downloaded.

```
private

def csv_name
   @csv_name ||= "guests-data-#{current_event.id}-#{Time.current.to_i}.csv
end

```
and then

```
file_name = Rails.root.join('tmp', csv_name)
...
send_file(file_name)

```
Check the x-send-file option and test it.

spec/interactors/spree_cm_commissioner/generate_guests_csv_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  let(:collection) { SpreeCmCommissioner::Guest.where(id: [guest1.id, guest2.id]) }
+  let(:file_name) { 'csv_file.csv' }
+  let(:csv_file) { CSV.read(Rails.root.join("tmp/#{file_name}")) }
+
+  describe ".call" do
+    it 'creates the CSV file with correct path and filename' do
+      expect(context).to be_a_success
+      expect(File.exist?(Rails.root.join("tmp/#{context.file_name}"))).to be_truthy
+      expect(context.file_name).to eq(file_name)
+    end
```
channainfo, line 18: Merge this into one only -- expect the file to be created correctly and read the content of the file back with correct data.

channainfo (changes requested):

app/controllers/spree/events/guests_controller.rb
vaneath: I'm not so sure how to implement it  bong, these are the resources that I'm been researching recently. But it works fine right now.

References:
https://api.rubyonrails.org/v7.1.3/classes/ActionController/DataStreaming.html#method-i-send_file
https://github.com/heroku/heroku-buildpack-ruby/issues/783
https://gist.github.com/djo/11374407

app/controllers/spree/events/guests_controller.rb
```diff
@@ -11,17 +11,35 @@ def model_class
         SpreeCmCommissioner::Guest
       end
 
+      def index
+        respond_with(collection) do |format|
+          format.csv do
+            file_name = Rails.root.join('tmp', csv_name)
```
channainfo, line 17: Move this out of the block

channainfo (changes requested):

PR #1214, opened 2024-03-14
close #1208 add seat selection availability flag on route

### LINKED ISSUE #1208: Implement seat selection availability flag on routes

tykealy:
**Added new column "allow_seat_selection" to cm_trips**

**quantity is required instead of seats when allow_seat_selection==false**

**seat layout  will be nil for trip that doesn't allow seat selection**

image

### DIFF
--- a/app/queries/spree_cm_commissioner/trip_seat_layout_query.rb
+++ b/app/queries/spree_cm_commissioner/trip_seat_layout_query.rb
@@ -10,35 +10,51 @@ module SpreeCmCommissioner
       @vehicle_type = SpreeCmCommissioner::VehicleType.find_by(id: vehicle.vehicle_type_id)
     end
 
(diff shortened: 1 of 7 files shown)


app/queries/spree_cm_commissioner/trip_seat_layout_query.rb
```diff
@@ -10,35 +10,45 @@ def initialize(trip_id:, date:)
...
-                            seat_id: seat.seat_id,
-                            vehicle_type_id: seat.vehicle_type_id
-                          }
-                        end
-                      end
-                    end
+      allow_seat_selection = @route.trip.allow_seat_selection
+
+      if allow_seat_selection
+        vehicle_seats = seats.to_a
```
channainfo, line 21: move the body of the condition to a separate method to reduce the number of lines in this method body.

channainfo (changes requested):

PR #1218, opened 2024-03-15
close #1217 filtered seats for order that is completed

### LINKED ISSUE #1217: add vehicle_seats filter for only completed order

tykealy:


db/migrate/20240315061802_add_state_index_to_spree_order.rb
```diff
@@ -0,0 +1,5 @@
+class AddStateIndexToSpreeOrder < ActiveRecord::Migration[7.0]
+  def change
+    add_index :spree_orders, :state, name: 'index_spree_orders_on_state'
```
channainfo, line 3: No need the index for the state field. Index only works well if it has many different value like id, email. The state won't be good for indexing.

channainfo (changes requested):

PR #1221, opened 2024-03-15
close #1216 allow Spree::Price to belong to polymophic instead of variant

### LINKED ISSUE #1216: Allow Spree::Price to belong to polymorphic instead of variant.

theachoem:
Spree::Price will be later used for other models like pricing rate. Following are prices after migration.

image

## Reference
- https://github.com/SpreeTravel/spree_travel_core/blob/master/app/models/spree/price_decorator.rb

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,9 +3,7 @@ name: Commissioner Gem
 
 on:
   pull_request:
-    branches:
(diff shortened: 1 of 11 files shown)


channainfo:
@theachoem, let's create a milestone branch and direct our pull requests to that milestone instead.

PR #1234, opened 2024-03-20
wip #1233 override variant option_values

### LINKED ISSUE #1233: override variant option_values

panhachom:
In App we want to return only option values that its option_type type is not hidden , In doing so we don't have modify our app code

### DIFF
--- a/app/serializers/spree/v2/storefront/variant_serializer_decorator.rb
+++ b/app/serializers/spree/v2/storefront/variant_serializer_decorator.rb
@@ -6,7 +6,10 @@ module Spree
           base.attributes :permanent_stock, :need_confirmation, :product_type, :kyc, :reminder_option_value, :started_at_option_value
 
           base.has_many :stock_locations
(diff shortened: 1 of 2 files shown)


app/serializers/spree/v2/storefront/variant_serializer_decorator.rb
```diff
@@ -6,7 +6,10 @@ def self.prepended(base)
           base.attributes :permanent_stock, :need_confirmation, :product_type, :kyc, :reminder_option_value, :started_at_option_value
 
           base.has_many :stock_locations
-          base.has_many :visible_option_values, serializer: Spree::V2::Storefront::OptionValueSerializer
+          # override
+          base.has_many :option_values do |object|
+            object.option_values.joins(:option_type).where(spree_option_types: { hidden: false })
```
channainfo, line 11: Create a custom relation instead. Avoid using any query here.

channainfo (changes requested):

PR #1287, opened 2024-03-30
close #1286 reassign event id for guest when assigned wrong

### LINKED ISSUE #1286: Reassign event id when wrong assign for guest

theachoem:
Make sure to re-assign event id even it is not null to ensure correct event id.

For example, we happen to assign wrong event id here. Event id should be 39 instead of 9. This will break check-in app.

image


theachoem:
image

PR #1302, opened 2024-04-04
Close #1281 Billing invoice not create bug

### LINKED ISSUE #1281: Billing invoice not create bug

Regene27:
image

### DIFF
--- a/app/models/spree_cm_commissioner/variant_decorator.rb
+++ b/app/models/spree_cm_commissioner/variant_decorator.rb
@@ -27,8 +27,8 @@ module SpreeCmCommissioner
 
     def display_variant
       display_sku = sku.split('-').join(' ')
-      display_price = price.to_i == price ? price.to_i : price
-      "#{display_sku} - #{display_price} ៛"
+      display_price = Spree::Money.new(price.to_i, currency: currency).to_s
+      "#{display_sku} - #{display_price}"
     end
 
     private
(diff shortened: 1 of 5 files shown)


app/models/spree_cm_commissioner/variant_decorator.rb
```diff
@@ -27,8 +27,9 @@ def selected_option_value_ids
 
     def display_variant
       display_sku = sku.split('-').join(' ')
-      display_price = price.to_i == price ? price.to_i : price
-      "#{display_sku} - #{display_price} ៛"
+      price.to_i == price ? price.to_i : price
```
channainfo, line 30: This line appears to have no effect. let's add some tests to clarify its intended purpose?

app/views/spree/billing/shared/_invoice.html.erb
```diff
@@ -26,6 +26,11 @@
   @invoice_number = @invoice.invoice_number
   @customer_number = @customer.customer_number
   @customer_address = @order.billing_address&.address1
+  if @current_vendor.payment_qrcode.present?
+    @qrcode = "<img src=\"#{main_app.rails_blob_url(@current_vendor.payment_qrcode.attachment)}\" alt=\"qrcode\" style=\"width: 150px;\">".html_safe
```
channainfo, line 30: use a helper instead.

channainfo (changes requested):

PR #1320, opened 2024-04-08
close #1313 setup pricing model/rate model migrations

### LINKED ISSUE #1313: Setup pricing rate/model database

theachoem:
pricing model rate (2)

## Introduce new tables
- cm_pricing_rates: belong to rateable (eg. variant)
- cm_pricing_models: belong to modelable (eg. variant)
- cm_pricing_rules: belong ruleable (in this case, it either belong to `cm_pricing_rates` or `cm_pricing_models`)
- cm_pricing_actions: belong `cm_pricing_models`
- cm_applied_pricing_models: created when model is applied.
   model cannot be deleted if there is a model applied.
- cm_applied_pricing_rates: created when rate is applied.
   rate cannot be deleted if there is a rate applied.

## New columns to line_items
- pricing_rates_amount
- pricing_models_amount
- pricing_subtotal (rate + model)

### DIFF
--- a/app/models/concerns/spree_cm_commissioner/line_item_pricings_concern.rb
+++ b/app/models/concerns/spree_cm_commissioner/line_item_pricings_concern.rb
@@ -1,0 +1,21 @@
+module SpreeCmCommissioner
+  module LineItemPricingsConcern
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :applied_pricing_rates, class_name: 'SpreeCmCommissioner::AppliedPricingRate', dependent: :destroy
+      has_many :applied_pricing_models, class_name: 'SpreeCmCommissioner::AppliedPricingModel', dependent: :destroy
+
+      has_many :pricing_rates, class_name: 'SpreeCmCommissioner::PricingRate', through: :applied_pricing_rates, source: :pricing_rate
+      has_many :pricing_models, class_name: 'SpreeCmCommissioner::PricingModel', through: :applied_pricing_models, source: :pricing_model
+
+      validates :pricing_models_amount, allow_nil: true, numericality: true
+      with_options allow_nil: true, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: Spree::Price::MAXIMUM_AMOUNT } do
+        validates :pricing_rates_amount
+        validates :pricing_subtotal
+      end
+
+      money_methods :pricing_rates_amount, :pricing_models_amount, :pricing_subtotal
+    end
+  end
+end
--- a/app/models/spree_cm_commissioner/price_decorator.rb
+++ b/app/models/spree_cm_commissioner/price_decorator.rb
@@ -4,6 +4,24 @@ module SpreeCmCommissioner
     def self.prepended(base)
       base.belongs_to :priceable, polymorphic: true, inverse_of: :prices, touch: true, optional: false
     end
+
+    def tax_category
+      priceable.respond_to?(:tax_category) ? priceable.tax_category : nil
+    end
+
+    # override:
+    # to get tax_category from priceable instead of variant
+    def price_including_vat_for(price_options)
+      options = price_options.merge(tax_category: tax_category)
+      gross_amount(price, options)
+    end
+
+    # override:
+    # to get tax_category from priceable instead of variant
+    def compare_at_price_including_vat_for(price_options)
+      options = price_options.merge(tax_category: tax_category)
+      gross_amount(compare_at_price, options)
+    end
   end
 end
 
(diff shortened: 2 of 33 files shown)


app/models/spree_cm_commissioner/price_decorator.rb
```diff
@@ -4,6 +4,22 @@ module PriceDecorator
...
+      gross_amount(price, options)
+    end
+
+    # override
+    def compare_at_price_including_vat_for(price_options)
+      options = price_options.merge(tax_category: tax_category)
+      gross_amount(compare_at_price, options)
+    end
   end
 end
```
theachoem, line 25: I changed from:

```rb
options = price_options.merge(tax_category: variant.tax_category)
```

```rb
options = price_options.merge(tax_category: tax_category)
```

app/models/concerns/spree_cm_commissioner/pricing_concern.rb
```diff
@@ -0,0 +1,33 @@
+# Module for pricing model & rate. It's expected to have following columns:
+# :match_policy, :effective_from, :effective_to, :priority
```
channainfo, line 2: if so, it would be better to remove the concern in favor of a plain ruby module instead.
I would call this module Ruleable or RuleablePrice instead.

app/models/spree_cm_commissioner/pricing_model.rb
```diff
@@ -0,0 +1,10 @@
+module SpreeCmCommissioner
+  class PricingModel < Base
+    include PricingConcern
+
+    belongs_to :modelable, optional: false, polymorphic: true, inverse_of: :pricing_models, touch: true
```
channainfo, line 5: This name might be taken by other gems. how about changing it to modelable_price instead?

app/models/concerns/spree_cm_commissioner/line_item_pricings_concern.rb
```diff
@@ -0,0 +1,18 @@
+module SpreeCmCommissioner
+  module LineItemPricingsConcern
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :applied_pricing_rates, class_name: 'SpreeCmCommissioner::AppliedPricingRate', dependent: :destroy
+      has_many :applied_pricing_models, class_name: 'SpreeCmCommissioner::AppliedPricingModel', dependent: :destroy
+
```
channainfo, line 8: add also has_many pricing_models / pricing_rates through applied_pricing_models / applied_pricing_rates

app/models/concerns/spree_cm_commissioner/priceable_concern.rb
```diff
@@ -11,6 +11,8 @@ def price_in(currency)
       cache_key = "spree/prices/#{cache_key_with_version}/price_in/#{currency}"
 
       Rails.cache.fetch(cache_key) { find_or_build_price(currency) } || find_or_build_price(currency)
+    rescue TypeError # when mocking in rspec, it raised TypeError: singleton can't be dumped
+      find_or_build_price(currency)
```
channainfo, line 15: let's check this work around

channainfo (changes requested):

app/models/concerns/spree_cm_commissioner/pricing_concern.rb
theachoem: https://api.rubyonrails.org/v7.1.3.2/classes/ActiveSupport/Concern.html.

concern look cleaner bong.

app/models/spree_cm_commissioner/pricing_model.rb
theachoem: fixed!

app/models/concerns/spree_cm_commissioner/line_item_pricings_concern.rb
theachoem: fixed!

app/models/concerns/spree_cm_commissioner/priceable_concern.rb
theachoem: Error happens because of cache & loaded prices.

```rb
def price_in(currency)
  Rails.cache.fetch(cache_key) do
    find_or_build_price(currency)
  end
end

def find_or_build_price(currency)
  if prices.loaded?
    prices.detect { |price| price.currency == currency }
  else
    prices.find_or_initialize_by(currency: currency)
  end
end
```

When prices is loaded, we find price via prices.detect. I assume that loaded price are loaded to be singleton. That why when try to cache it, it raised `singleton can't be dumped` which mean we can't cache singleton class.

To fix it, we can just change find_or_build_price method to like following:
```rb
def price_in(currency)
  Rails.cache.fetch(cache_key) do
    find_or_build_price(currency)
  end
end

def find_or_build_price(currency)
   prices.find_or_initialize_by(currency: currency)
end
```

PR #1367, opened 2024-04-29
 close #1366 review images api

panhachom:
### UI Design 
image

### DIFF
--- a/app/controllers/spree/api/v2/storefront/review_images_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/review_images_controller.rb
@@ -1,0 +1,31 @@
+module Spree
+  module Api
+    module V2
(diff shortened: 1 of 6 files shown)


app/controllers/spree/api/v2/storefront/review_images_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+            if context.success?
+              render_serialized_payload { serialize_resource(context.result) }
+            else
+              render_error_payload(context.message)
+            end
+          end
+
+          private
+
+          def load_review
```
channainfo, line 22: rename from load_review to review instead.

channainfo (changes requested):

PR #1382, opened 2024-05-03
close #1381 create review selectors with spree taxon

### LINKED ISSUE #1381: create review selectors with taxon

panhachom:
## Design 
test

## Backend Set up 
image

image

image

### DIFF
--- a/app/controllers/spree/admin/taxon_star_ratings_controller.rb
+++ b/app/controllers/spree/admin/taxon_star_ratings_controller.rb
@@ -1,0 +1,31 @@
+module Spree
+  module Admin
+    class TaxonStarRatingsController < Spree::Admin::ResourceController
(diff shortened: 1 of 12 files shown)


app/controllers/spree/admin/taxons_controller_decorator.rb
```diff
@@ -23,6 +26,10 @@ def remove_home_banner
 
       private
 
+      def set_rating
+        @taxon.subtitle = calculate_rating_value(params[:taxon][:subtitle]) if @taxon.review?
```
channainfo, line 30: create a new table: taxon_review_star and add necessary fields and logic to that table instead to prevent confusion in the future.

channainfo (changes requested):

PR #1392, opened 2024-05-06
close #1372 change issued-date from created_at to start_date

### LINKED ISSUE #1372: Invoice issue_date need to be service start_date not invoice created_at date
The reason for the change is to show the service start_date so that when the order/invoice recurs, the date will be consistent.

tykealy:
The reason for the change is to show the service start_date so that when the order/invoice recurs, the date will be consistent.

### DIFF
--- a/app/views/spree/billing/report/balance_due.html.erb
+++ b/app/views/spree/billing/report/balance_due.html.erb
@@ -50,7 +50,7 @@
         <% @orders.each do |order| %>
           <tr id="<%= spree_dom_id order %>" data-hook="admin_subscriptions_index_rows">
             <td><%= link_to order.invoice&.invoice_number, edit_billing_order_path(order) %></td>
-            <td><%= order.created_at.to_date %></td>
+            <td><%= order.line_items[0].from_date.to_date %></td>
             <td>
               <% if order.customer.first_name.present? %>
                 <%= link_to order.customer.first_name, edit_billing_customer_path(order.customer) %>
--- a/app/views/spree/billing/report/paid.html.erb
+++ b/app/views/spree/billing/report/paid.html.erb
@@ -38,7 +38,7 @@
         <% @orders.each do |order| %>
           <tr id="<%= spree_dom_id order %>" data-hook="admin_subscriptions_index_rows">
             <td><%= link_to order.invoice&.invoice_number, edit_billing_order_path(order) %></td>
-            <td><%= order.created_at.to_date %></td>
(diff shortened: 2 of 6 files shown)


channainfo:
@tykealy plz add the reason why we did this in the description to we can learn and wont repeat this in the future

Regene27:
The reason for the change is to show the service start_date so that when the order/invoice recurs, the date will be consistent.

channainfo:
> The reason for the change is to show the service start_date so that when the order/invoice recurs, the date will be consistent.

Put in the description of the issue as I mentioned above.

app/views/spree/billing/report/balance_due.html.erb
```diff
@@ -50,7 +50,7 @@
         <% @orders.each do |order| %>
           <tr id="<%= spree_dom_id order %>" data-hook="admin_subscriptions_index_rows">
             <td><%= link_to order.invoice&.invoice_number, edit_billing_order_path(order) %></td>
-            <td><%= order.created_at.to_date %></td>
+            <td><%= order.line_items[0].from_date.to_date %></td>
```
channainfo, line 53: make sure you include line_items in the like this .includes(:line_items) to avoid n+1 query

app/views/spree/billing/report/paid.html.erb
```diff
@@ -38,7 +38,7 @@
         <% @orders.each do |order| %>
           <tr id="<%= spree_dom_id order %>" data-hook="admin_subscriptions_index_rows">
             <td><%= link_to order.invoice&.invoice_number, edit_billing_order_path(order) %></td>
-            <td><%= order.created_at.to_date %></td>
+            <td><%= order.line_items[0].from_date.to_date %></td>
```
channainfo, line 41: make sure you include line_items in the like this .includes(:line_items) to avoid n+1 query

app/views/spree/billing/report/show.html.erb
```diff
@@ -27,7 +27,7 @@
         <% @orders.each do |order| %>
           <tr id="<%= spree_dom_id order %>" data-hook="admin_subscriptions_index_rows">
             <td><%= link_to order.invoice&.invoice_number, edit_billing_order_path(order) %></td>
-            <td><%= order.created_at.to_date %></td>
+            <td><%= order.line_items[0].from_date.to_date %></td>
```
channainfo, line 30: make sure you include line_items in the like this .includes(:line_items) to avoid n+1 query

channainfo (changes requested):

PR #1452, opened 2024-05-28
Close #1448 Filter out subscribed variants

### LINKED ISSUE #1448: filter out subscribed variant

Regene27:
image
image

- Hide all variants of a product when a variant of that product is subscribed


app/models/spree_cm_commissioner/customer.rb
```diff
@@ -89,6 +89,22 @@ def query_builder
                                        .joins('INNER JOIN spree_line_items as li ON li.order_id = o.id ')
     end
 
+    def filtered_variants
+      if subscriptions.present?
+        subscribable_variants
```
channainfo, line 94: Let's return all the variant and the variants that have been subscribed. You can calculate. You can calculate that via Ruby instead of doing it all with SQL for simplicity:

Show something like this in the UI:

Variant a - ( subscribed )
Variant b - ( subscribed )
Variant c
Variant d

The example above the customer have already subscribe to  Variant a and Variant b

channainfo (changes requested):

app/models/spree_cm_commissioner/customer.rb
```diff
@@ -89,6 +89,14 @@ def query_builder
                                        .joins('INNER JOIN spree_line_items as li ON li.order_id = o.id ')
     end
 
+    def subscribed_variants
+      subscribed_variants = []
+      subscriptions.each do |subscription|
+        subscribed_variants << subscription.variant
+      end
+      subscribed_variants
```
channainfo, line 97: No need this method, you can use this in short: 

```
subscriptions.map(&:variant)
```

or to get just the id

```
subscriptions.map(&:variant_id)

```

app/views/spree/billing/subscriptions/new.html.erb
```diff
@@ -7,6 +7,33 @@
   <%= render partial: 'spree/admin/shared/error_messages', locals: { target: @object } %>
 </div>
 
+<% if @subscribed_variants.any? %>
+  <div class="table-responsive border rounded bg-white mb-3">
```
channainfo, line 11: Move this to a partial

app/views/spree/billing/subscriptions/new.html.erb
```diff
@@ -7,6 +7,33 @@
...
+      <thead class="text-muted">
+        <th><%= Spree.t(:subscribed_variants) %></th>
+        <th><%= Spree.t(:variant_quantity) %></th>
+        <th><%= Spree.t(:subscribed_date) %></th>
+      </thead>
+      <tbody>
+        <% @subscribed_variants.each do |subscription| %>
+          <tr>
+            <td>
+              <li><%= subscription.variant.display_variant %></li>
```
channainfo, line 22: includes(:variant)

channainfo (changes requested):

PR #1473, opened 2024-05-30
close #1462 manually create new invoice

### LINKED ISSUE #1462: Allow customer to subscribe more than one variant of a product

tykealy:
- Allow multiple variants of a product to be subscribed
- Customer has many orders/invoices
- Invoices/Orders tab in Customer Tabs
- Add button to manually create new Order/Invoice
- When creating new order/invoice, if ( currrent_month == last_invoice.month ) ==> FAIL
image


app/controllers/spree/billing/customers_controller.rb
```diff
@@ -21,6 +21,18 @@ def load_bussinesses
         @businesses = Spree::Taxonomy.businesses.taxons.where('depth > ? ', 1).order('parent_id ASC').uniq
       end
```
channainfo, line 23: Add comment on the  with Http method and route path

app/interactors/spree_cm_commissioner/subscriptions_order_creator.rb
```diff
@@ -0,0 +1,50 @@
...
+    def create_order
+      active_subscriptions = context.customer.active_subscriptions
+      return if active_subscriptions.blank?
+
+      context.new_order = customer.orders.create!(
+        subscription_id: active_subscriptions.first.id,
+        phone_number: context.customer.phone_number,
+        user_id: context.customer.user_id
+      )
+
```
channainfo, line 28: use today = Time.zone.today and then reuse it instead.

db/migrate/20240530041030_add_customer_cm_orders.rb
```diff
@@ -0,0 +1,5 @@
+class AddCustomerCmOrders < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_orders, :customer_id, :integer, if_not_exists: true
```
channainfo, line 3: let's discuss about this.

channainfo (changes requested):

app/controllers/spree/billing/customers_controller.rb
channainfo: # GET /path

db/migrate/20240530041030_add_customer_cm_orders.rb
channainfo: Generate a user with email: admin-{SecureRedom.hex(16)}@samram.com, pwd: SecureRedom.hex(32)

channainfo (changes requested):

PR #1509, opened 2024-06-06
close #1508 refactor option type/value - use name as value instead of presentation

### LINKED ISSUE #1508: Improve option type/value - use name as value instead of presentation
The presentation should be displayed to the user. When we use it as value, it's not understandable.
- After this change, we have to rename some of option value manually
- We have to update app as app use to read presentation and decode to value instead of name.

theachoem:


theachoem:
# Documentation
## I. Reserved Options

There are some name that used in our system. In that case, we disable name from editing.

image

## II. Attr Types

Following are current available attr_types & its UI:

### Float

image

### Integer

image

### String

image

### Boolean

image

### Date

image

### Time

image

### Coordinate

image

### State Selection

image

### Payment option

image

### Delivery Option

image

PR #1515, opened 2024-06-07
close #1471 add video upload field to S3

### LINKED ISSUE #1471: Add video upload field to S3
image
image

SreyMochOrng:
## Video upload under Variant ( one variant has one video )

 - video tab
 
image

### DIFF
--- a/app/models/spree_cm_commissioner/video_on_demand.rb
+++ b/app/models/spree_cm_commissioner/video_on_demand.rb
@@ -1,0 +1,10 @@
+module SpreeCmCommissioner
+  class VideoOnDemand < SpreeCmCommissioner::Base
+    belongs_to :variant, class_name: 'Spree::Variant'
+    has_one_attached :file
+    has_one_attached :thumbnail
+
+    validates :title, :description, :file, :thumbnail, presence: true
+    validates :variant_id, uniqueness: true
+  end
+end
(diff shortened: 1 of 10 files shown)


app/models/spree_cm_commissioner/video_on_demand.rb
```diff
@@ -0,0 +1,10 @@
+module SpreeCmCommissioner
+  class VideoOnDemand < SpreeCmCommissioner::Base
+    belongs_to :variant, class_name: 'Spree::Variant'
+    has_one_attached :clip
```
channainfo, line 4: let's rename from **clip** to **file**.

channainfo (changes requested):

app/models/spree_cm_commissioner/video_on_demand.rb
SreyMochOrng: Done bong

app/overrides/spree/admin/shared/_product_tabs/video_on_demand.html.erb.deface
```diff
@@ -0,0 +1,9 @@
+<!-- insert_after "erb[silent]:contains('Spree::Digital')" -->
+<% if @product.product_type == 'video' %>
```
channainfo, line 2: Let's remove this condition all kinds of products can upload video.

channainfo (changes requested):

PR #1517, opened 2024-06-10
close #1516 added invoice and order recurring

### LINKED ISSUE #1516: Invoice and Order Recurring

tykealy:
***Future Subscription***
- When a subscription is scheduled to start in the future (i.e., next month or beyond), an invoice will not be generated immediately.
- The invoice for the subscription will be generated automatically when the current month matches the subscription's start month.

***Missing Invoices***
- To address potential invoice creation failures, a new field, last_invoice_date, has been added to the cm_customers table.
- This field will be used to track the date of the last successful invoice creation.
- If there are gaps in the invoicing period (i.e., months where invoices were not generated), the system will calculate the number of missing invoices based on the last_invoice_date and create the necessary invoices accordingly.

### DIFF
--- a/app/interactors/spree_cm_commissioner/subscribed_order_creator.rb
+++ b/app/interactors/spree_cm_commissioner/subscribed_order_creator.rb
@@ -4,15 +4,21 @@ module SpreeCmCommissioner
 
     def call
       # reject if locked?
+      today = Time.zone.today
+      start_date = subscription.start_date
+      return if  start_date.beginning_of_month > today.beginning_of_month
       return unless subscription.update(last_occurence: subscription.renewal_date)
 
       find_or_create_order
+      set_last_invoice_date
       create_line_item
       create_invoice
 
       context.order.create_default_payment_if_eligble
       context.order.reload
     end
+
+    private
 
     def find_or_create_order
       customer = subscription.customer
@@ -46,6 +52,13 @@ module SpreeCmCommissioner
           }
         )
       end
+    end
+
+    def set_last_invoice_date
+      last_invoice_date = context.subscription.customer.last_invoice_date
+      return if last_invoice_date.present? && last_invoice_date.beginning_of_month > Time.zone.now.beginning_of_month
+
+      context.subscription.customer.last_invoice_date = Time.zone.now
     end
 
     def renewal_date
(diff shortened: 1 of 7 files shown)


app/interactors/spree_cm_commissioner/subscribed_order_creator.rb
```diff
@@ -4,9 +4,12 @@ class SubscribedOrderCreator < BaseInteractor
 
     def call
       # reject if locked?
-      return unless subscription.update(last_occurence: subscription.renewal_date)
+      today = Time.zone.today
+      start_date = subscription.start_date
+      return if  start_date.beginning_of_month > today.beginning_of_month || !subscription.update(last_occurence: subscription.renewal_date)
```
channainfo, line 9: let's this into a separate statement

```
return if  start_date.beginning_of_month > today.beginning_of_month
return if !subscription.update(last_occurence: subscription.renewal_date)
```


```

File: app/interactors/spree_cm_commissioner/subscribed_order_creator.rb [hunk 2]
```diff
@@ -48,6 +51,14 @@ def create_line_item
       end
     end
 
+    def set_last_invoice_date
+      last_invoice_date = context.subscription.customer.last_invoice_date
+
+      return unless last_invoice_date.nil? || last_invoice_date.beginning_of_month < Time.zone.now.beginning_of_month
```
channainfo, line 57: break this into a separate statement

app/interactors/spree_cm_commissioner/subscriptions_order_creator.rb
```diff
@@ -1,27 +1,38 @@
...
-      else
-        create_order
+        return
+      end
+
+      return unless last_invoice_date.beginning_of_month < today
+
+      missed_months = missed_months(last_invoice_date.beginning_of_month, today)
+      missed_months.times do |i|
+        month = i + 1
```
channainfo, line 18: move the body of this block to a separate method. Never run big logic in the loop

channainfo (changes requested):

channainfo:
@tykealy after refactoring as suggested above, let's add more test to cover different cases

PR #1524, opened 2024-06-12
close #1523 create thumbnail model for video on demand

### LINKED ISSUE #1523: create thumbnail model for video on-demand

SreyMochOrng:
## Create Thumbnail model for video on demand
- beside call it _"has_on_attached: :thumbnail"_

image

## Why change to this?
- Because when create new Thumbnail model it generate url when like image


app/models/spree_cm_commissioner/thumbnail.rb
```diff
@@ -0,0 +1,11 @@
+module SpreeCmCommissioner
+  class Thumbnail < SpreeCmCommissioner::Asset
+    def asset_styles
+      {
+        mini: '187x67>',
```
channainfo, line 5: Thumbnails suppose to have the same dimension as video

channainfo (changes requested):

app/models/spree_cm_commissioner/video_on_demand_thumbnail.rb
```diff
@@ -0,0 +1,11 @@
+module SpreeCmCommissioner
+  class VideoOnDemandThumbnail < SpreeCmCommissioner::Asset
+    def asset_styles
+      {
+        mini: '160x90>',
+        small: '240x135>',
+        large: '360x202>'
```
channainfo, line 7: the the same size as video's.

channainfo (changes requested):

PR #1541, opened 2024-06-19
close #1540 added current invoice indicator

### LINKED ISSUE #1540: Add current month invoice indicator

tykealy:
-add current month's invoice indicator
image

- pagination for customers and products.


channainfo:
@tykealy you better do it on the server side code instead.

PR #1564, opened 2024-06-25
close #1563 add confirm pin code checker

### LINKED ISSUE #1563: Add confirm pin code checker

vaneath:
## App Demo: https://github.com/bookmebus/cm-market-app/pull/1449

```ruby
def has_secure_password(attribute = :password, validations: true)
        # Load bcrypt gem only when has_secure_password is used.
        # This is to avoid ActiveModel (and by extension the entire framework)
        # being dependent on a binary library.
        begin
          require "bcrypt"
        rescue LoadError
          $stderr.puts "You don't have bcrypt installed in your application. Please add it to your Gemfile and run bundle install"
          raise
        end

        include InstanceMethodsOnActivation.new(attribute)

        if validations
          include ActiveModel::Validations

          # This ensures the model has a password by checking whether the password_digest
          # is present, so that this works with both new and existing records. However,
          # when there is an error, the message is added to the password attribute instead
          # so that the error message will make sense to the end-user.
          validate do |record|
            record.errors.add(attribute, :blank) unless record.public_send("#{attribute}_digest").present?
          end

          validates_length_of attribute, maximum: ActiveModel::SecurePassword::MAX_PASSWORD_LENGTH_ALLOWED
          validates_confirmation_of attribute, allow_blank: true
        end
      end
```

```ruby
 class InstanceMethodsOnActivation < Module
      def initialize(attribute)
        attr_reader attribute

        define_method("#{attribute}=") do |unencrypted_password|
          if unencrypted_password.nil?
            instance_variable_set("@#{attribute}", nil)
            self.public_send("#{attribute}_digest=", nil)
          elsif !unencrypted_password.empty?
            instance_variable_set("@#{attribute}", unencrypted_password)
            cost = ActiveModel::SecurePassword.min_cost ? BCrypt::Engine::MIN_COST : BCrypt::Engine.cost
            self.public_send("#{attribute}_digest=", BCrypt::Password.create(unencrypted_password, cost: cost))
          end
        end

        define_method("#{attribute}_confirmation=") do |unencrypted_password|
          instance_variable_set("@#{attribute}_confirmation", unencrypted_password)
        end

        # Returns +self+ if the password is correct, otherwise +false+.
        #
        #   class User < ActiveRecord::Base
        #     has_secure_password validations: false
        #   end
        #
        #   user = User.new(name: 'david', password: 'mUc3m00RsqyRe')
        #   user.save
        #   user.authenticate_password('notright')      # => false
        #   user.authenticate_password('mUc3m00RsqyRe') # => user
        define_method("authenticate_#{attribute}") do |unencrypted_password|
          attribute_digest = public_send("#{attribute}_digest")
          attribute_digest.present? && BCrypt::Password.new(attribute_digest).is_password?(unencrypted_password) && self
        end

        alias_method :authenticate, :authenticate_password if attribute == :password
      end
    end
```

### DIFF
--- a/app/controllers/spree/api/v2/storefront/confirm_pin_code_checkers_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/confirm_pin_code_checkers_controller.rb
@@ -1,0 +1,24 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class ConfirmPinCodeCheckersController < ::Spree::Api::V2::ResourceController
+          before_action :require_spree_current_user
+
+          def update
+            context = SpreeCmCommissioner::ConfirmPinCodeChecker.call(
+              input_pin_code: params[:pin_code],
+              user_pin_code: spree_current_user.confirm_pin_code_digest
+            )
+
+            if context.success?
+              render_serialized_payload { { status: 'ok' } }
+            else
+              render_error_payload(context.message, 400)
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/interactors/spree_cm_commissioner/confirm_pin_code_checker.rb
+++ b/app/interactors/spree_cm_commissioner/confirm_pin_code_checker.rb
@@ -1,0 +1,17 @@
+module SpreeCmCommissioner
+  class ConfirmPinCodeChecker < BaseInteractor
+    delegate :input_pin_code, :user_pin_code, to: :context
+
+    def call
+      context.fail!(message: :pin_code_must_not_blank) if input_pin_code.blank?
+
+      validate_pin_code!
+    end
+
+    private
+
+    def validate_pin_code!
+      context.fail!(message: :invalid_pin_code) unless BCrypt::Password.new(user_pin_code) == input_pin_code
+    end
+  end
+end
--- a/app/models/spree_cm_commissioner/user_decorator.rb
+++ b/app/models/spree_cm_commissioner/user_decorator.rb
@@ -7,8 +7,6 @@ module SpreeCmCommissioner
 
       base.enum gender: %i[male female other]
 
-      base.encrypts :encrypted_confirm_pin_code
-
       base.has_many :subscriptions, through: :customer, class_name: 'SpreeCmCommissioner::Subscription'
       base.has_many :payments, as: :payable, class_name: 'Spree::Payment', dependent: :nullify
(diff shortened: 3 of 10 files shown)


app/controllers/spree/api/v2/storefront/confirm_pin_code_checkers_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+          before_action :require_spree_current_user
+
+          def update
+            context = SpreeCmCommissioner::ConfirmPinCodeChecker.call(
+              input_pin_code: params[:pin_code],
+              user_pin_code: spree_current_user.confirm_pin_code_digest
+            )
+
+            if context.success?
+              render_serialized_payload { { status: 'ok' } }
```
channainfo, line 15: To create a generic serializer object that represents the required fields (id, type, and status), we can use a consistent response payload. This ensures that the client library can adapt without having to handle each case separately.

example

```
# app/serializers/generic_serializer.rb
class GenericSerializer < ActiveModel::Serializer
  attributes :id, :type, :status

  def id
     SecureRandom.hex(16)
  end
end
```

channainfo (changes requested):

app/controllers/spree/api/v2/storefront/confirm_pin_code_checkers_controller.rb
vaneath: Thanks bong, I follow from the existing code in our commissioner and it uses ```spree_api``` BaseController for render payload.

```ruby
def render_serialized_payload(status = 200)
  render json: yield, status: status, content_type: content_type
end

def render_error_payload(error, status = 422)
  json = if error.is_a?(ActiveModel::Errors)
            { error: error.full_messages.to_sentence, errors: error.messages }
          elsif error.is_a?(Struct)
            { error: error.to_s, errors: error.to_h }
          else
            { error: error }
          end

  render json: json, status: status, content_type: content_type
end

def render_result(result, ok_status = 200)
  if result.success?
    render_serialized_payload(ok_status) { serialize_resource(result.value) }
  else
    render_error_payload(result.error)
  end
end
```

So, I should custom generic serializer to render payload for it right bong? To consistant response payload

channainfo: yes, just define a serialize resource to the generic one if the response is not related to any specific model.

PR #1593, opened 2024-07-03
close #1591 fix bug device token and Refactor Notification

### LINKED ISSUE #1591: Fix bug Database store Multiple DeviceToken for User

LengTech11:
## Fixed in this PR
- [x]  Fix if it delete and register device token or not.
- [x]  Fix notification why it sent notification multiple time for scheduled CustomerNotification
- [x]  Create delete duplicate device token in SideKiq.
- [x]  Check Schedule Send if it have sent_at do not let it send.
- [x]  For Sent_all Check Notification if Recipient already received make sure it not push to the SameUser Again.
- [x] Add Possible Test Case for Push Customer Notification Flow


app/interactors/spree_cm_commissioner/customer_notification_sender.rb
```diff
@@ -3,20 +3,22 @@ class CustomerNotificationSender < BaseInteractor
     delegate :user_ids, :customer_notification, to: :context
 
     def call
+      update_sent_at
       send_notification
```
channainfo, line 7: let's send the notification before update the table to avoid a prematura update coz if the send notification fails then we would have a false positive sent_at.

app/interactors/spree_cm_commissioner/customer_notification_sender.rb
```diff
@@ -53,17 +43,26 @@ def deliver_notification_to_user(user)
...
-      Spree.user_class.end_users_push_notificable.joins(
-        "LEFT JOIN cm_notifications ON (
-          cm_notifications.recipient_id = spree_users.id
-          AND cm_notifications.notificable_type = 'SpreeCmCommissioner::CustomerNotification'
-          AND cm_notifications.notificable_id = #{customer_notification.id}
-        )"
-      )
+      if customer_notification.notification_taxons.any?
+        SpreeCmCommissioner::UsersByEventFetcherQuery
+          .new(customer_notification.notification_taxons.pluck(:taxon_id))
```
channainfo, line 60: Do not use this ustomer_notification.notification_taxons.pluck(:taxon_id) because if the dataset is big we will face the problem.

channainfo (changes requested):

app/interactors/spree_cm_commissioner/customer_notification_sender.rb
LengTech11: I use select with find_each and batch_size 1000.

PR #1644, opened 2024-07-12
close #1643 auto create option value when editing variant

### LINKED ISSUE #1643: Auto create option value when input from variant

theachoem:
After this, admin doesn't have to go to option type to pre-create option values.

| |
| - |
| image |
| image |


channainfo:
@theachoem, do we need to migrate existing data to accommodate these changes?

PR #1646, opened 2024-07-12
Close #1623 Billing load confirmation popup when capture or void payment

### LINKED ISSUE #1623: billing load confirmation pop up when capture payment

Regene27:
https://github.com/user-attachments/assets/26667ca5-af6b-4848-af07-eb4bf2e531a0

- Show pop up window for confirmation
- Add reason if order is void


db/migrate/20240710061915_add_reason_to_spree_orders.rb
```diff
@@ -0,0 +1,5 @@
+class AddReasonToSpreeOrders < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_orders, :reason, :text, if_not_exists: true
```
channainfo, line 3: we already have fields like internal_note, special_instructions, ... I don't see why you need to create a new field for this.

channainfo (changes requested):

PR #1667, opened 2024-07-18
close #1632 generate bib number

### LINKED ISSUE #1632: Generate BIB number

vaneath:
[Screencast from 2024-07-22 17-30-03.webm](https://github.com/user-attachments/assets/8c9a511d-6ddd-4817-9a52-3c340f04fb4e)

### DIFF
--- a/app/models/concerns/spree_cm_commissioner/bib_number_concern.rb
+++ b/app/models/concerns/spree_cm_commissioner/bib_number_concern.rb
@@ -1,0 +1,35 @@
+module SpreeCmCommissioner
+  module BibNumberConcern
+    extend ActiveSupport::Concern
+
+    included do
+      after_update :generate_bib_number_job, if: :should_generate_bib_number?
+
+      def generate_bib_number_job
+        # Iterate over the filtered line items
+        line_items_with_bib_number_prefix_and_kyc.each do |line_item|
+          line_item.guests.where(bib_number: [nil, '']).find_each(&:generate_bib_number)
+        end
+      end
+
+      private
+
+      # Check if the state has changed to complete and at least one line item has a bib_number_prefix
+      def should_generate_bib_number?
+        state_changed_to_complete? && bib_number_prefix?
+      end
+
+      # Verify if any line item has a bib_number_prefix
+      def bib_number_prefix?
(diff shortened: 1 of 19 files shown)


app/jobs/spree_cm_commissioner/generate_bib_number_job.rb
```diff
@@ -0,0 +1,13 @@
+module SpreeCmCommissioner
+  class GenerateBibNumberJob < ApplicationJob
+    def perform(taxon_id)
+      taxon = Spree::Taxon.find(taxon_id)
```
channainfo, line 4: Move the logic to BibNumberGenerator class and rename this to BibNumberGeneratorJob

channainfo (changes requested):

channainfo:
The bib number and the trail should be stored as meta fields in the line items. Storing the bib in the guest is not ideal because not all products require a bib. I think we might not need the bib number auto generated as well.

app/models/concerns/spree_cm_commissioner/bib_number_concern.rb
```diff
@@ -0,0 +1,15 @@
+module SpreeCmCommissioner
+  module BibNumberConcern
+    extend ActiveSupport::Concern
+
+    included do
+      after_update :generate_bib_number_job, if: -> { state_changed_to_complete? }
```
channainfo, line 6: Let's generate bib number only if the product bib prefix is present

db/migrate/20240726043406_add_partial_unique_index_to_bib_number_and_bib_number_prefix.rb
```diff
@@ -0,0 +1,8 @@
+class AddPartialUniqueIndexToBibNumberAndBibNumberPrefix < ActiveRecord::Migration[7.0]
+  def change
+    add_index :cm_guests, [:bib_number, :bib_number_prefix, :event_id],
```
channainfo, line 3: Let's remove this for now. Create a new ticket for this. The code here is complicated

theachoem:
If we used this https://github.com/channainfo/commissioner/pull/1705 instead, let's close this PR as duplicate.

vaneath:
Move to #1705

PR #1674, opened 2024-07-19
close #1599 change invoice recurring logic

### LINKED ISSUE #1599: change invoicing logic

tykealy:
**New Invoice Logic:**
- only create invoice if the service has been used for at least a month

Example:
- User subscribe to a service on **June 15th 2024**
- The invoice won't be generated immediately like the previous logic
- The invoice will only be generated on or after **July 15th 2024**

Invoice coverage
- invoice that is from **15th May** to **14th June** is considered to be June's invoice.
- invoice that is form **15th June** to **14th July** is considered to be July's invoice.


spec/interactors/spree_cm_commissioner/subscriptions_order_creator_spec.rb
```diff
@@ -27,14 +27,18 @@
     context "when customer has subscriptions" do
       context "when there are 3 missed months" do
         before do
-          create(:cm_subscription, customer: customer, quantity: 2, start_date: 3.months.ago.beginning_of_month)
-          create(:cm_subscription, customer: customer, quantity: 2, start_date: 3.months.ago.beginning_of_month)
-          customer.last_invoice_date =  3.months.ago
+          allow_any_instance_of(SpreeCmCommissioner::Subscription).to receive(:date_within_range).and_return(true)
+          today = Time.zone.today
```
channainfo, line 31: Use the travel_to method to stub the today value so that you can have different scenario

channainfo (changes requested):

PR #1691, opened 2024-07-24
close #1678 implement video required on product

### LINKED ISSUE #1678: implement video required on product
- Add check box on product ( if check user need to upload video default )

image

 - List video with position is movable 
 
image

- Update some relationship at from one variant have only one video ==> one variant can has many video
- Create default variant for default video in case require_default_video is checked on product (which use product master_id instead of variant_id inside video_on_demand_table)

image

SreyMochOrng:
- Add check box on product ( if check user need to upload video default )

image

- when require video is checked and default not uploaded yet 

image

 - Video list with position is movable 
 
image

- Update some relationship at from one variant have only one video ==> one variant can has many video
- Create default variant for default video in case require_default_video is checked on product (which use product master_id instead of variant_id inside video_on_demand_table)

image

### DIFF
--- a/app/controllers/spree/admin/video_on_demands_controller.rb
+++ b/app/controllers/spree/admin/video_on_demands_controller.rb
@@ -1,7 +1,7 @@
 module Spree
   module Admin
     class VideoOnDemandsController < Spree::Admin::ResourceController
-      before_action :load_parent
+      before_action :product
       before_action :load_video_on_demands, only: %i[index new edit create]
 
       def new
@@ -35,6 +35,13 @@ module Spree
         end
       end
 
+      def update_positions
+        params[:positions].each do |id, index|
+          video_on_demand = SpreeCmCommissioner::VideoOnDemand.find(id)
+          video_on_demand.update(position: index)
+        end
+      end
+
       private
 
       def model_class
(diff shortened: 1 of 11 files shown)


db/migrate/20240722035047_add_require_video_to_spree_products.rb
```diff
@@ -0,0 +1,5 @@
+class AddRequireVideoToSpreeProducts < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_products, :require_video, :boolean, index: true, default: false, if_not_exists: true
```
channainfo, line 3: remove index

channainfo (changes requested):

db/migrate/20240722035047_add_require_video_to_spree_products.rb
```diff
@@ -0,0 +1,5 @@
+class AddRequireVideoToSpreeProducts < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_products, :require_video, :boolean, default: false, if_not_exists: true
```
channainfo, line 3: Rename this to use_video_as_default instead.

channainfo (changes requested):

channainfo:
Add this field use_video_as_default to the serializer as well

PR #1705, opened 2024-07-29
 close #1704 generate unique guest bib number in event scope 

### LINKED ISSUE #1704: Unique guest bib number in event scope

vaneath:


app/interactors/spree_cm_commissioner/bib_number_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    private
+
+    def taxon
+      @taxon ||= Spree::Taxon.find(taxon_id)
+    end
+
+    def guests_with_bib_number_prefix
+      taxon.guests.includes(:line_item, line_item: :option_types).where(bib_number: [nil, '']).select do |guest|
+        guest.line_item.option_types.any? { |option_type| option_type.name == 'bib-number-prefix' }
```
channainfo, line 17: let's extract the 
```sh
option_type.name == 'bib-number-prefix'

```
to its own method in the option_type_decorator.rb

```rb
def bib_number_prefix?
  name == 'bib-number-prefix'
end




```

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -112,5 +115,30 @@ def current_age
 
       ((Time.zone.now - dob.to_time) / 1.year.seconds).floor
     end
+
+    def set_bib_number_prefix
+      self.bib_number_prefix = bib_number_prefix
```
channainfo, line 120: this is confusing. The default bib_number_prefix should only be set once.

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -112,5 +115,30 @@ def current_age
...
+                             .where(bib_number_prefix: bib_number_prefix)
+                             .maximum(:bib_number)
+
+      sequence_number = last_bib_number ? last_bib_number.to_i + 1 : 1
+
+      self.bib_number = format('%03d', sequence_number)
+      save
+    end
+
+    def bib_number_prefix
```
channainfo, line 136: remove this method it causes confusion.

channainfo (changes requested):

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -112,5 +116,36 @@ def current_age
 
       ((Time.zone.now - dob.to_time) / 1.year.seconds).floor
     end
+
+    def set_bib_number_prefix
+      self.bib_number_prefix = line_item.variant.bib_number_prefix
+    end
+
+    def set_bib_index
+      return if bib_index.present? && !bib_related_fields_changed?
```
channainfo, line 125: let's remove this line. the logic here might bring us a false positive result

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -112,5 +116,36 @@ def current_age
...
+
+      self.bib_index = "#{event_id}-#{bib_number_prefix}-#{bib_number}" if bib_number.present?
+    end
+
+    def bib_related_fields_changed?
```
channainfo, line 130: let's remove this. field_changed? is problematic.

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -112,5 +116,36 @@ def current_age
...
     end
+
+    def set_bib_number_prefix
+      self.bib_number_prefix = line_item.variant.bib_number_prefix
+    end
+
+    def set_bib_index
+      return if bib_index.present? && !bib_related_fields_changed?
+
+      self.bib_index = "#{event_id}-#{bib_number_prefix}-#{bib_number}" if bib_number.present?
```
channainfo, line 127: If the bib_number field is blank, it should be generated and assigned before saving. We permit this field to be blank because there are occasions when we want to manually assign a value. A more effective approach might involve checking for the presence of bib_number_prefix.

If bib_number_prefix is not blank:

1. Generate bib_number if it is blank.
2. Assign the bib_index using the prefix and the number.

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -112,5 +116,36 @@ def current_age
...
+    end
+
+    def bib_related_fields_changed?
+      bib_number_changed? || bib_number_prefix_changed? || event_id_changed?
+    end
+
+    def generate_bib_number
```
channainfo, line 134: create two method generate_bib_number and generate_bib_number!

generate_bib_number return the next bib_number 

and 
```rb
def generate_bib_number!
  bib_number = generate_bib_number
  save!
end
```

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -112,5 +116,36 @@ def current_age
...
+                             .where(bib_number_prefix: bib_number_prefix)
+                             .maximum(:bib_number)
+
+      sequence_number = last_bib_number ? last_bib_number.to_i + 1 : 1
+
+      self.bib_number = format('%03d', sequence_number)
+      save
+    end
+
+    def full_bib_number
```
channainfo, line 147: use bib_number_with_prefix instead.

app/models/spree_cm_commissioner/option_type_decorator.rb
```diff
@@ -12,6 +12,10 @@ def self.prepended(base)
 
       base.scope :promoted, -> { where(promoted: true) }
       base.whitelisted_ransackable_attributes = %w[kind]
+
+      def bib_number_prefix?
```
channainfo, line 16: Maybe bib_prefix sounds more consistent

channainfo (changes requested):

app/models/spree_cm_commissioner/guest.rb
vaneath: Hi bong, thank for reviewing. But I am not quite sure what you mean here, I have resolved most of your comment, but I am still misunderstanding with this. According to what I know, bib_number should only be generated when conditions are meet. Example: kyc is enable and the option_type has bib-prefix-number, otherwise it will be blank. Could you me the example of can I implement this bong?

channainfo: rename bib_number_prefix to bib_prefix

channainfo (changes requested):

app/interactors/spree_cm_commissioner/bib_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+    private
+
+    def taxon
+      return @taxon if defined?(@taxon)
+
+      @taxon ||= Spree::Taxon.event.find_by(id: taxon_id)
+    end
+
+    def generate_guests_bib
+      return unless taxon
```
channainfo, line 18: let's do this instead
```
context.fail!(message: "invalid taxon event with id: #{taxon_id}") if taxon.blank?
```

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -18,6 +18,7 @@ class Guest < SpreeCmCommissioner::Base
 
     scope :complete, -> { joins(:line_item).merge(Spree::LineItem.complete) }
     scope :unassigned_event, -> { where(event_id: nil) }
+    scope :none_bib, -> { where(bib_prefix: [nil, '']).order(created_at: :asc) }
```
channainfo, line 21: let's remove order(created_at: :asc) from the scope. we should call the order  in the service instead.

app/interactors/spree_cm_commissioner/bib_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      taxon.guests.includes(line_item: :variant).none_bib.find_each(&:generate_bib!)
```
channainfo, line 20: # order the record by id is more efficient.
taxon.guests.includes(line_item: :variant).none_bib.order('id ASC').find_each(&:generate_bib!)

channainfo (changes requested):

app/models/spree_cm_commissioner/guest.rb
```diff
@@ -113,5 +121,33 @@ def current_age
...
+      self.bib_index = "#{event_id}-#{bib_prefix}-#{bib_number}"
+    end
+
+    def generate_bib!
+      generate_bib
+      save!
+    end
+
+    def bib_number_with_prefix
+      "#{bib_prefix}#{bib_number}"
```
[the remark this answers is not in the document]
theachoem: I think solution_2 is better for us, since we want to use them as seat as well. But we have to aware of its CONS.

theachoem (changes requested):

app/models/spree_cm_commissioner/guest.rb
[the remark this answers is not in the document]
vaneath: Noted, brother

PR #1710, opened 2024-07-30
Close #1709 super_admin role create migration

### LINKED ISSUE #1709: super_admin create migration

Regene27:


channainfo:
Do this in rails console instead.

PR #1762, opened 2024-08-13
close #1761 allow non-admin to delete promotion

### LINKED ISSUE #1761: add non-admin to delete promotion

tykealy:


channainfo:
@tykealy let's check the integrity of the data. What if the promotion has already been applied to the order and you delete the promotion? I would prefer to deactivate it if possible to avoid data integrity issue.

tykealy:
> @tykealy, let's verify the integrity of the data. What happens if a promotion that has already been applied to an order is deleted? To avoid issues with data integrity, I would suggest deactivating the promotion instead of deleting it if it has already been used.

@channainfo, Bong the promotion needs to be deleted to stop it from applying on the future recurring orders since we check if there is any promotion code matches the customer's ID to apply it every month. For existing orders that has been applied, the adjustment will be intact when the promotion is deleted.

PR #1780, opened 2024-08-15
close #1779 Remove .complete from guest generate bib number

### LINKED ISSUE #1779: Remove .complete from guest
Let say they buy 3 tickets = 3 guests.

last_bib_number = 23

The first guest is now generate the bib_number and get 24
The second guest is also generate the bib_number after first guest, but since order is not yet completed so first guest is not in complete scope then it will get 24 again.

So remove .complete will make more sense in this case.

vaneath:


channainfo:
@vaneath plz check the test failed.

vaneath:
> @vaneath plz check the test failed.

Hi bong, the failed test is related to other file which is not expected to happen, related to billing. I try to revert my code back and pull all the latest from develop, copy migration and re-run the test_app but still having that issues.

image

PR #1788, opened 2024-08-16
close #1787 send order details to telegram web app user when order completed

### LINKED ISSUE #1787: Send order details to telegram web app user when order completed

theachoem:
Telegram user ID is a chat ID, so we can just send them directly.

image


channainfo:
@theachoem, could you check the test failure? If it's the same issue as the one documented here https://github.com/channainfo/commissioner/pull/1780#issuecomment-2292803108, please comment it out for now until Tykea is available to fix it.

theachoem:
@channainfo yes bong! I will also add some tests

PR #1791, opened 2024-08-17
Close #1790 fix error with background image upload in guest card classes page

### LINKED ISSUE #1790: Fix error with background image upload in GuestCardClasses page
pic1

Sreyleak-Deth:
### Before:

pic1

### After:
image


app/controllers/spree/admin/guest_card_classes_controller.rb
```diff
@@ -52,10 +60,13 @@ def location_after_save
 
       # override
       # permit all attributes for now.
+
       def permitted_resource_params
-        key = ActiveModel::Naming.param_key(@object)
-        permit_keys = params.require(key).keys
-        params.require(key).permit(permit_keys)
+        @permitted_resource_params ||= begin
+          key = ActiveModel::Naming.param_key(@object)
```
channainfo, line 66: Let's permit the params keys explicitly one by one here.

channainfo (changes requested):

app/controllers/spree/admin/guest_card_classes_controller.rb
Sreyleak-Deth: > Let's permit the params keys explicitly one by one here.

Got it bong, I will update it, Thanks

PR #1799, opened 2024-08-20
Close #1796 allow users to download imported orders csv

### LINKED ISSUE #1796: Allow Users to Download Imported Orders CSV

panhachom:
https://github.com/user-attachments/assets/1cd854f3-c5d5-4407-84da-52a5c37b8bd2

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -183,8 +183,16 @@ Spree::Core::Engine.add_routes do
 
     resources :orders, except: [:show] do
       collection do
-        resources :import_existing_orders
-        resources :import_new_orders
+        resources :import_existing_orders do
+          member do
+            get :download
+          end
+        end
+        resources :import_new_orders do
+          member do
+            get :download
+          end
+        end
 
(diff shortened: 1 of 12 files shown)


config/routes.rb
```diff
@@ -183,8 +183,16 @@
...
+            get :download_imported_file
+          end
+        end
+        resources :import_new_orders do
+          member do
+            get :download_imported_file
+          end
+        end
 
         get :download_new_order_csv_template, controller: 'import_new_orders'
```
channainfo, line 197: Place this under the resources block to ensure that actions are easily visible and rename it to just "download"

```
resources  : import_new_orders do
   member do
       get :download
   end
end
```

config/routes.rb
```diff
@@ -183,8 +183,16 @@
...
         get :download_existing_order_csv_template, controller: 'import_existing_orders'
```
channainfo, line 198: Here too.

channainfo (changes requested):

PR #1827, opened 2024-08-23
close #1826 support multiple telegram web bot

### LINKED ISSUE #1826: Support multiple telegram web bots

theachoem:
### 1. New telegram_bot MVC

| |
| - |
| a. Index |
| image |
| b. New |
| image |
| c. Edit, we hide token here for security reasons, admin can read actual tokens from the database or @bot_father directly. |
| image |

### 2. Make sure it create association between user identity provider & telegram when open bot
- Refactor UserTelegramWebAppAuthenticator for validate initData with telegram web bot from new model (TelegramBot).
  Client must pass tg_bot when authenticate in advance.
- Refactor OrderCompleteTelegramSender to send message to user with bot from new model instead of bot

> So we no longer need default telegram bot from env for Telegram web app related features

DEMO:

### DIFF
--- a/app/interactors/spree_cm_commissioner/user_telegram_web_app_authenticator.rb
+++ b/app/interactors/spree_cm_commissioner/user_telegram_web_app_authenticator.rb
@@ -1,16 +1,20 @@
 module SpreeCmCommissioner
   class UserTelegramWebAppAuthenticator < BaseInteractor
-    delegate :telegram_init_data, to: :context
+    delegate :telegram_init_data, :telegram_bot_username, to: :context
 
     def call
-      return context.fail!(message: 'invalid_init_data') unless checker_context.success?
+      return context.fail!(message: 'telegram_bot_username_is_not_valid_or_not_exist_in_db') if telegram_bot.blank?
+      return context.fail!(message: 'invalid_init_data') if checker.blank?
 
-      context.telegram_user = ActiveSupport::JSON.decode(checker_context.decoded_telegram_init_data['user'])
+      context.telegram_user = ActiveSupport::JSON.decode(checker.checker_context.decoded_telegram_init_data['user'])
       context.user = find_or_create_telegram_user
     end
 
     def find_or_create_telegram_user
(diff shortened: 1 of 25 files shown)


app/interactors/spree_cm_commissioner/user_telegram_web_app_authenticator.rb
```diff
@@ -20,22 +24,32 @@ def find_or_create_telegram_user
...
       end
     end
 
-    def checker_context
-      context.checker_context ||= SpreeCmCommissioner::TelegramWebAppInitDataValidator.call(
-        telegram_init_data: telegram_init_data,
-        bot_token: telegram_bot_token
-      )
+    def checker
+      # currently, when open telegram web app, telegram did not pass bot username for us,
```
channainfo, line 37: When linking the app with Telegram, we can include our registered bot ID in the query string like this:
https://www.bookmeplus.com/t/events/xx?tgb_id=our_registered_bot_id

Then, we can extract the our_registered_bot_id value from the tgb_id parameter and use it to find the corresponding bot token in our database.

If the tgb_id is not present we can just assume the bot is our bot ( the default one). What do you think?

channainfo (changes requested):

app/interactors/spree_cm_commissioner/user_telegram_web_app_authenticator.rb
theachoem: We can do that

db/migrate/20240823093714_create_cm_telegram_bots.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCmTelegramBots < ActiveRecord::Migration[7.0]
+  def change
+    create_table :cm_telegram_bots, if_not_exists: true do |t|
+      t.string :username
```
channainfo, line 4: let's add an index to the username  column to efficiently query bot my username.

I think we also should add an index to the token column as well  so in the future we might want to allow the admin to search bot my token.

channainfo (changes requested):

PR #1833, opened 2024-08-26
close #1832 add address model, serilizer decorator

### LINKED ISSUE #1832: Add AddressDecorator to prepend enum of gender

vaneath:


app/models/spree_cm_commissioner/address_decorator.rb
```diff
@@ -0,0 +1,9 @@
+module SpreeCmCommissioner
+  module AddressDecorator
+    def self.prepended(base)
+      base.enum gender: { :other => 0, :male => 1, :female => 2 }
```
channainfo, line 4: let's do this:
{ not_selected: 0, male: 1, female: 2, other: 3 }. In the UI if use the not_selected as default.

channainfo (changes requested):

app/models/spree_cm_commissioner/address_decorator.rb
vaneath: Field is not mandatory for contact information. Use 
```ruby
enum gender: { :none => 0, :male => 1, :female => 2, :other => 3 }
```

PR #1845, opened 2024-08-28
close #1844 add seat mapper option type & its generator tools

### LINKED ISSUE #1844: Add seat number mapper option type & generator tools rb

theachoem:
## In this PR
1. Add new reserve option type 'seat-number-positions' with new attr_type `array`
2. Add seat_number columns to the guest (no index or validation added)
3. Add a seat generator tool for solo concerts specifically.

image


channainfo:
Storing seat labels has become complicated due to their connection to a physical seat inventory. To make things simpler, we can map the seats during check-in based on the actual guest count. This approach might affect offline as well.

theachoem:
That would be another flow that hasn't been discussed yet. With this flow, users don't know their seat until check in. 

As discussed, we want to display seats to them. We can use default spree variant stock for managing stock. And HM & us will recheck stock everyday.

theachoem:
That would be another flow that hasn't been discussed yet. With this flow, users don't know their seat until check in. 

As discussed, we want to display seats to them. We can use default spree variant stock for managing stock. And HM & us will recheck stock everyday. 

Can you check and improve this one instead?

theachoem:
### Limitation
When an order is complete, it generates a bib number (eg. VIP20) which then generates a seat at position 20.
If the order is canceled, or marked as not paid, the seat will still remain the same. Bib number will continue to get bigger which leaves the seat unoccupied.

In this case, simple solution is to add those seats to suffixes to allow other users to buy it. Example:
- A11
- A12
- A13
- A14 <-- this seat is canceled.
- A15
- A16
- A14 <-- add A14 to last, so it can be purchase again.

PR #1910, opened 2024-09-09
close #1909 query for active customer for invoice recurring only

### LINKED ISSUE #1906: Add customer deactivation

### LINKED ISSUE #1909: only query for active customer for invoice recurring

tykealy:
- **close #1906 added customer deactivation**

- when customer is switched from `inactive` to `active`, update customer's last invoice date to `today`.
- query only active customer to perform invoice recurring.

### DIFF
--- a/app/interactors/spree_cm_commissioner/subscriptions_order_cron_executor.rb
+++ b/app/interactors/spree_cm_commissioner/subscriptions_order_cron_executor.rb
@@ -2,7 +2,7 @@ module SpreeCmCommissioner
   class SubscriptionsOrderCronExecutor < BaseInteractor
     def call
       today = Time.zone.today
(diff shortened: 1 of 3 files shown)


db/migrate/20240909051440_add_status_to_cm_customers.rb
```diff
@@ -0,0 +1,6 @@
+class AddStatusToCmCustomers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :cm_customers, :status, :integer, null: false, default: 0, if_not_exists: true
+    add_index :cm_customers, :status, if_not_exists: true
```
channainfo, line 4: let's remove the index. index does not help much in this case.

channainfo (changes requested):

PR #1964, opened 2024-09-19
Close #1963 Change the status from show_event_status to show_badge_status

### LINKED ISSUE #1963: Change the status from event_status to badge_status

DethSreyleak:
- app : https://github.com/bookmebus/cm-market-app/pull/1794

Screenshot from 2024-09-20 09-25-18
Screenshot from 2024-09-20 09-25-26

### For the UI in the app, I talked with Bong Hour. He said there is no need to change the UI; we should just keep the old UI because the color has meaning for the status. I tested placing the show_badge_status in the top right corner, but it didn't look good, so I decided to keep it like the old UI as well.

image

image


db/migrate/20240907061117_add_show_event_status_to_spree_taxon.rb
```diff
@@ -1,5 +0,0 @@
-class AddShowEventStatusToSpreeTaxon < ActiveRecord::Migration[7.0]
-  def change
-    add_column :spree_taxons, :show_event_status, :boolean, if_not_exists: true
```
channainfo, line 3: Don't delete this file

db/migrate/20240907061117_add_badge_status_to_spree_taxon.rb
```diff
@@ -0,0 +1,5 @@
+class AddBadgeStatusToSpreeTaxon < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_taxons, :badge_status, :boolean, if_not_exists: true
```
channainfo, line 3: use the rename column instead.  let's call it show_badge_status and default to the false value

app/overrides/spree/admin/taxons/_form/badge_status.html.erb.deface
```diff
@@ -0,0 +1,11 @@
+<!-- insert_before "erb[loud]:contains('field_container :parent_id')" -->
+
+<%= f.field_container :badge_status do %>
+  <%= f.label :badge_status, Spree.t('badge_status'), class: 'form-label' %>
+  <%= f.select :badge_status,
+        options_for_select([['On', true], ['Off', false]], f.object.badge_status),
+        { include_blank: 'None' },
```
channainfo, line 7: remove include blank. by default is Off ( false)

channainfo (changes requested):

app/overrides/spree/admin/taxons/_form/badge_status.html.erb.deface
channainfo: @Sreyleak-Deth make sure you test db:migrate and db:rollback to ensure they work correctly

Sreyleak-Deth:
Note bong, I will check it again

db/migrate/20240907061117_add_show_badge_status_to_spree_taxon.rb
```diff
@@ -0,0 +1,5 @@
+class AddShowBadgeStatusToSpreeTaxon < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_taxons, :show_badge_status, :boolean, default: false, if_not_exists: true
```
channainfo, line 3: We already have a column for this, so why not rename the existing column instead of creating a new one?

channainfo (changes requested):

db/migrate/20240907061117_add_show_badge_status_to_spree_taxon.rb
Sreyleak-Deth: > We already have a column for this, so why not rename the existing column instead of creating a new one?

The original column was named show_event_status, and based on your suggestion, I have renamed it to show_badge_status.

db/migrate/20240926035819_rename_show_event_status_to_show_badge_status_to_spree_taxon.rb
```diff
@@ -0,0 +1,6 @@
+class RenameShowEventStatusToShowBadgeStatusToSpreeTaxon < ActiveRecord::Migration[7.0]
+  def change
+    rename_column :spree_taxons, :show_event_status, :show_badge_status
+    change_column :spree_taxons, :show_badge_status, :boolean, default: false, if_not_exists: true
```
channainfo, line 4: remove the if_not_exists: true.
```
change_column :spree_taxons, :show_badge_status, :boolean, default: false
```

channainfo (changes requested):

db/migrate/20240926035819_rename_show_event_status_to_show_badge_status_to_spree_taxon.rb
Sreyleak-Deth: > remove the if_not_exists: true.
> 
> ```
> change_column :spree_taxons, :show_badge_status, :boolean, default: false
> ```

it has fixed bong

PR #1976, opened 2024-09-25
close #1975 Fix kyc resource not found

### LINKED ISSUE #1975: Fix Kyc resource not found

SreyMochOrng:
### Fix error on kyc struggling on loading when editing 

image

### DIFF
--- a/app/controllers/spree/api/v2/storefront/guests_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/guests_controller.rb
@@ -13,17 +13,6 @@ module Spree
             model_class.all
           end
 
-          # show by guest token
-          def show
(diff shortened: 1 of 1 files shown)


app/controllers/spree/api/v2/storefront/guests_controller.rb
```diff
@@ -13,17 +13,6 @@ def collection
             model_class.all
           end
 
-          # show by guest token
-          def show
-            guest = model_class.find_by(token: params[:id])
```
channainfo, line 18: Are we using this API to query the guest in the HM ticket system? Another solution for compatibility could be checking the length of params[:id] to determine whether to search by ID or token.

channainfo (changes requested):

PR #1998, opened 2024-10-08
close #1997 recheck line item stock

### LINKED ISSUE #1997: Recheck Line item stock

panhachom:
### App Demo
https://github.com/bookmebus/cm-market-app/pull/1813


db/migrate/20241008025609_add_stock_available_to_spree_line_item.rb
```diff
@@ -0,0 +1,5 @@
+class AddStockAvailableToSpreeLineItem < ActiveRecord::Migration[7.0]
+  def change
+    add_column :spree_line_items, :stock_available, :boolean, default: true, if_not_exists: true
```
channainfo, line 3: I don't think it's necessary to store this. If the stock is replenished (either new items are added or cancelled tickets are returned), the stock_available value might become outdated and will need to be synchronized.

channainfo (changes requested):

PR #2037, opened 2024-10-31
Close #2033 enhance BookMe+ order confirmation

### LINKED ISSUE #2033: Enhance BookMe+ order confirmation
image

SreyMochOrng:
## New Email Order Confirmation 
image


channainfo:
we need to generate a more obscure and sharable link instead

PR #2046, opened 2024-11-06
WIP#2045 add nearby places suggestion

nangdypanhar:


nangdypanhar:
Screenshot : 
image

theachoem (approved):

app/controllers/spree/api/v2/platform/places_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+          def index
+            result = collection_serializer.new(collection).serializable_hash
+
+            respond_with(result) do |format|
+              format.json { render json: result }
+            end
+          end
+
+          def collection
+            @collection ||= begin
```
channainfo, line 17: remove the **begin** statement here cause nested code. let's remove it.

app/controllers/spree/api/v2/platform/places_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+            SpreeCmCommissioner::Api::V2::Platform::PlaceSerializer
+          end
+
+          def collection_serializer
+            SpreeCmCommissioner::Api::V2::Platform::PlaceSerializer
+          end
+
+          private
+
+          Place = Struct.new(:id, :name, :base_64_content)
```
channainfo, line 45: Remove this Struct from global declaration in the controller, it might cause a conflict of type in the future.

app/interactors/spree_cm_commissioner/google_places_fetcher.rb
```diff
@@ -0,0 +1,41 @@
+module SpreeCmCommissioner
+  class GooglePlacesFetcher < BaseInteractor
+    delegate :name, to: :context
+
+    def call
+      context.google_places = []
+      json_response = fetch_google_place_by_name
+
+      json_response['results'].each_with_index do |json, _|
```
channainfo, line 9: use the .map instead.

app/interactors/spree_cm_commissioner/google_places_fetcher.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    private
+
+    def google_map_key
+      @google_map_key ||= ENV.fetch('GOOGLE_MAP_KEY')
+    end
+
+    def fetch_google_place_by_name
+      url =   URI("#{GOOGLE_MAP_API_URL}/place/textsearch/json?query=#{name}&key=#{google_map_key}")
```
channainfo, line 32: Don't construct the query manually, it is not safe.
```
queries = {
   query: name,
   key: google_map_key
}.to_param

google_uri =   URI("#{GOOGLE_MAP_API_URL}/place/textsearch/json?#{queries}")

```

app/interactors/spree_cm_commissioner/google_places_fetcher.rb
```diff
@@ -0,0 +1,41 @@
...
+      https = Net::HTTP.new(url.host, url.port)
```
channainfo, line 33: Use the **Faraday** instead of the **Net**. Make sure to keep the Http  outbound  request consitent by not employing other libs/classes

app/controllers/spree/api/v2/platform/places_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+          def struct_places_with_id(places)
+            places.map do |place|
+              Place.new(
+                id: place.reference,
+                name: place.name,
+                base_64_content: Base64.strict_encode64(place.to_json)
```
channainfo, line 51: I forget. why do we need the base_64_content for?

channainfo (changes requested):

app/controllers/spree/api/v2/platform/places_controller.rb
nangdypanhar: In form, we can only pass key & value to it bong. Because we want to pass all google place data, we need to encode it.

In controller, we can decode it back, and create place in database if not yet exist.

PR #2048, opened 2024-11-07
Close #2047 implement JWT on email confirmation

### LINKED ISSUE #2047: Implement JWT on email confirmation

SreyMochOrng:
|When we click <b>View More</b>, it will redirect to web booking details|it redirect to this page without authentication by just using JWT token to verify|
|-|-|
|image|image|

### DIFF
--- a/app/controllers/spree/api/v2/storefront/anonymous_order_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/anonymous_order_controller.rb
@@ -1,0 +1,41 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class AnonymousOrderController < Spree::Api::V2::BaseController
+          def show_anonymous_order
+            token = params[:token]
+            order = order_jwt_token(token)
+
+            if order
+              render_serialized_payload { serialize_resource(order) }
+            else
+              render json: { error: 'Invalid or expired token' }, status: :unauthorized
+            end
+          end
+
+          def resource_serializer
+            Spree::V2::Storefront::OrderSerializer
+          end
+
+          private
+
+          def order_jwt_token(token)
+            decoded_token = SpreeCmCommissioner::OrderJwtToken.decode(token)
+
+            order_number = decoded_token['order_number']
+            return nil unless order_number
+
+            order = Spree::Order.find_by(number: order_number)
+            return nil unless order
+
+            decoded_token = SpreeCmCommissioner::OrderJwtToken.decode(token, order&.token)
+            return nil unless decoded_token
+
+            order
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


app/controllers/spree/api/v2/storefront/anonymous_order_controller.rb
```diff
@@ -0,0 +1,34 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class AnonymousOrderController < Spree::Api::V2::BaseController
+
+          def show_anonymous_order
+            token = params[:token]
+
+            decoded_token = JWT.decode(token, nil, false)
```
channainfo, line 10: Let's encapsulate this to class to find an order by jwt_token called order_jwt_token ( encode, decode ).

channainfo (changes requested):

app/controllers/spree/api/v2/storefront/anonymous_order_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+            end
+          end
+
+          def resource_serializer
+            Spree::V2::Storefront::OrderSerializer
+          end
+
+          private
+
+          def order_jwt_token(token)
```
channainfo, line 23: Let's move this to libs/cm_commissioners/order_jwt_token.rb service

```
module CmCommissioner
  class OrderJwtToken
      def self.encode(order)
           # construct the payload to encode here
           # encode  and return jwt_token string
      end

      def self.decode(jwt_token)
           # decode and return order or nil is invalid
      end
   end
end






```

app/controllers/spree/api/v2/storefront/anonymous_order_controller.rb
```diff
@@ -0,0 +1,47 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        class AnonymousOrderController < Spree::Api::V2::BaseController
+          def show_anonymous_order
+            token = params[:token]
+            order = order_jwt_token(token)
```
channainfo, line 8: order = CmCommissioner::OrderJwtToken.decode(token)

app/mailers/spree/order_mailer_decorator.rb
```diff
@@ -18,6 +18,8 @@ def confirm_email(order, resend: false)
 
       subject = (resend ? "[#{Spree.t(:resend).upcase}] " : '')
       subject += "#{@current_store&.name} Booking Confirmation ##{@order.number}"
+      payload = { order_number: @order.number, user_id: @order.user.id, store_id: @current_store.id }
+      @jwt_token = JWT.encode(payload, @order.token, 'HS256')
```
channainfo, line 22: @jwt_token = OrderJwtToken.encode(@order)

app/mailers/spree/order_mailer_decorator.rb
```diff
@@ -18,6 +18,8 @@ def confirm_email(order, resend: false)
 
       subject = (resend ? "[#{Spree.t(:resend).upcase}] " : '')
       subject += "#{@current_store&.name} Booking Confirmation ##{@order.number}"
+      payload = { order_number: @order.number, user_id: @order.user.id, store_id: @current_store.id }
```
channainfo, line 21: use **@order.user_id** instead of **@order.user.id**

channainfo (changes requested):

PR #2059, opened 2024-11-18
close #2058 invalidate cache for frequently use API

### LINKED ISSUE #2058: Invalidate cache for frequently used API

theachoem:
## In this PR:
1. Invalidate cache
image

2. Set cache to header for GET or HEAD. eg. `content.bookme.plus`

Since we use MiniProfilerRails, we can't set cache control at all. It will override.

| MiniProfilerRails | MiniProfilerRails (I disabled it for test) |
| - | - |
| image | image |

Related PR:
- Changes on app to request to CDN for frequently use API: https://github.com/bookmebus/cm-market-app/pull/1915
- Changes on client to request to CDN for frequently use API: https://github.com/bookmebus/cm-market-client/pull/267

### DIFF
--- a/app/controllers/concerns/spree_cm_commissioner/content_cachable.rb
+++ b/app/controllers/concerns/spree_cm_commissioner/content_cachable.rb
@@ -1,0 +1,20 @@
+module SpreeCmCommissioner
+  module ContentCachable
+    extend ActiveSupport::Concern
+
+    included do
+      after_action :set_cache_control_for_cdn
+    end
+
+    def max_age
+      ENV.fetch('CONTENT_CACHE_MAX_AGE', '7200')
+    end
+
+    def set_cache_control_for_cdn
+      return unless request.get? || request.head?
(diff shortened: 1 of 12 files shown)


app/controllers/concerns/spree_cm_commissioner/content_cachable.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    included do
+      after_action :set_cache_control_for_cdn
+    end
+
+    def set_cache_control_for_cdn
+      return unless request.get? || request.head?
+      return unless request.base_url == ENV['CONTENT_HOST_URL']
+
+      response.set_header('Cache-Control', 'public, max-age=7200')
```
channainfo, line 13: Let's make the 7200 in a ENV['CACHE_MAX_AGE']. and default it to 7200 if it does not exist.

channainfo (changes requested):

PR #2086, opened 2024-11-25
close #2080 fix bug taxon reposition not work

### LINKED ISSUE #2080: fix reposition taxon not work

LengTech11:
### Description

When adding the `has_many :children` association to the `Taxon` model in `SpreeCmCommissioner::TaxonDecorator`, taxon repositioning fails. The `children` association is ordered by the `position` field, which conflicts with **Awesome Nested Set**'s use of `lft` and `rgt` for hierarchical relationships.

### Steps to Reproduce

1. Add `SpreeCmCommissioner::TaxonDecorator` to `Spree::Taxon`.
2. Create a hierarchy of taxons with children.
3. Try to reorder taxons or change a taxon’s parent.
4. Observe that repositioning fails because `position` is used instead of `lft`/`rgt`.

### Expected Behavior

Repositioning should update the taxon's position based on the `lft` and `rgt` columns as used by **Awesome Nested Set**.

### Actual Behavior

Repositioning does not work properly because taxons are ordered by the `position` field.

### Fix

1. Remove reliance on the `position` field.
2. Order `children` by `lft` for **Awesome Nested Set** compatibility.
3. Add helper methods to manage repositioning using **Awesome Nested Set**.

### Solution

1. **Update the `children` association**:
   - Use `lft` to order children.

2. **Add helper methods** to reposition taxons in the hierarchy.

```ruby
module SpreeCmCommissioner
  module TaxonDecorator
    def self.prepended(base)

      # Update `children` to order by `lft`
      base.has_many :children, -> { order(:lft) }, class_name: 'Spree::Taxon', foreign_key: :parent_id, dependent: :destroy

  end
end

Spree::Taxon.prepend(SpreeCmCommissioner::TaxonDecorator) unless Spree::Taxon.included_modules.include?(SpreeCmCommissioner::TaxonDecorator)

### DIFF
--- a/app/models/spree_cm_commissioner/taxon_decorator.rb
+++ b/app/models/spree_cm_commissioner/taxon_decorator.rb
@@ -27,7 +27,8 @@ module SpreeCmCommissioner
       base.has_one :app_banner, as: :viewable, dependent: :destroy, class_name: 'SpreeCmCommissioner::TaxonAppBanner'
       base.has_one :home_banner, as: :viewable, dependent: :destroy, class_name: 'SpreeCmCommissioner::TaxonHomeBanner'
 
-      base.has_many :children, class_name: 'Spree::Taxon', foreign_key: :parent_id, dependent: :destroy
+      # Update children association to work with nested set (lft, rgt)
+      base.has_many :children, -> { order(:lft) }, class_name: 'Spree::Taxon', foreign_key: :parent_id, dependent: :destroy
       base.has_many :children_classifications, through: :children, source: :classifications, class_name: 'Spree::Classification'
 
       base.has_many :notification_taxons, class_name: 'SpreeCmCommissioner::NotificationTaxon'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/models/spree_cm_commissioner/taxon_decorator.rb [hunk 1]
```diff
@@ -27,7 +27,8 @@ def self.prepended(base)
       base.has_one :app_banner, as: :viewable, dependent: :destroy, class_name: 'SpreeCmCommissioner::TaxonAppBanner'
       base.has_one :home_banner, as: :viewable, dependent: :destroy, class_name: 'SpreeCmCommissioner::TaxonHomeBanner'
 
-      base.has_many :children, class_name: 'Spree::Taxon', foreign_key: :parent_id, dependent: :destroy
+      # Update children association to work with nested set (lft, rgt)
+      base.has_many :children, -> { order(:lft) }, class_name: 'Spree::Taxon', foreign_key: :parent_id, dependent: :destroy
```
channainfo, line 31: let's add a spec for this.

channainfo (changes requested):

PR #2113, opened 2024-12-05
close #2112 enqueue add to cart

### LINKED ISSUE #2112: create job when user add to cart

LengTech11:
Changes on app:
https://github.com/bookmebus/cm-market-app/pull/2006

image

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -391,6 +391,10 @@ Spree::Core::Engine.add_routes do
           patch :restart_checkout_flow
         end
 
+        namespace :queue_cart do
+          resources :line_items, only: [:create]
+        end
+
         resources :wished_items
         resources :user_promotion
         resources :order_promotions
--- a/app/controllers/spree/api/v2/storefront/queue_cart/line_items_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/queue_cart/line_items_controller.rb
@@ -1,0 +1,56 @@
+module Spree
+  module Api
+    module V2
+      module Storefront
+        module QueueCart
+          class LineItemsController < CartController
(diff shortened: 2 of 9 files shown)


theachoem:
@LengTech11

Panha will take over this.

app/controllers/spree/api/v2/storefront/enqueue_cart_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+            result = SpreeCmCommissioner::EnqueueCart::AddItemStatusMarker.call(
+              order_number: spree_current_order.number,
+              job_id: job.job_id,
+              status: 'processing',
+              queued_at: Time.current,
+              variant_id: @variant.id,
+              quantity: add_item_params[:quantity]
+            )
+
+            struct = Struct.new(:id, :status, :queued_at, :collection_reference, :document_reference)
```
channainfo, line 29: let's move this to a class ( serializable_objects )

config/routes.rb
```diff
@@ -391,6 +391,9 @@
           patch :restart_checkout_flow
         end
 
+        resource :enqueue_cart, controller: :enqueue_cart do
```
channainfo, line 394: resources :add_item_queues, only: [:create]

to avoid a custom action like this

```
# POST /add_item
  def add_item
  end
end
```

channainfo (changes requested):

config/routes.rb
panhachom:
```
  namespace :queue_cart do
          resources :line_items, only: [:create]
  end
  ```
 how about i contruct like this bong ? We will use create for add_item and destroy for remove item

PR #2136, opened 2024-12-11
close #2135-improve-firestore-connection

### LINKED ISSUE #2135: Improve firebase connection
```
 def firestore
    @firestore ||= Google::Cloud::Firestore.new(project_id: service_account[:project_id], credentials: service_account)
  end
```

panhachom:


config/initializers/firestore_client.rb
```diff
@@ -0,0 +1,14 @@
+require 'google/cloud/firestore'
+
+class FirestoreClient
+  def self.instance
+    @instance ||= Google::Cloud::Firestore.new(
```
channainfo, line 5: This one might not be a thread-safe

channainfo (changes requested):

config/initializers/firestore_client.rb
channainfo: The implementation looks promissing but calling .instance is not thread-safed.

PR #2177, opened 2024-12-20
close #2156: lock db when generate bib_index to fix concurrency

### LINKED ISSUE #2156: ActiveRecord::RecordNotUnique: index_cm_guests_on_bib_index
[ActiveRecord::RecordNotUnique](https://bookmeplus.sentry.io/issues/6142761804/?project=4508452610506752&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=3)

PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_cm_guests_on_bib_index" (ActiveRecord::RecordNotUnique)
DETAIL:  Key (bib_index)=(478-VIP-138) already exists.

**PG::UniqueViolation**
ERROR:  duplicate key value violates unique constraint "index_cm_guests_on_bib_index" (PG::UniqueViolation)
DETAIL:  Key (bib_index)=(478-VIP-138) already exists.

**Note:**
1) Check if the process is auto generate the bib number, we can apply queue method.
2) If it is selected by user, we can apply pub/sub(when someone is selected the item, then it notify other) method with firebase.

hengsokly:
- When payment is `completed`, the order will update the state to `complete` and trigger guest to `generate_bib!` (`SpreeCmCommissioner::OrderBibNumberConcern`)
- The bib_index is validated as unique
- Users may make payments simultaneously that make the guest_bib_index duplicate and raise errors.
- If applying a queue, it may cause a delay and need to update the UI to handle the delay.
- So we apply lock db with pessimistic lock.

### DIFF
--- a/app/models/spree_cm_commissioner/guest.rb
+++ b/app/models/spree_cm_commissioner/guest.rb
@@ -180,23 +180,22 @@ module SpreeCmCommissioner
     end
 
     def generate_bib
-      return if bib_prefix.present?
-      return unless bib_required?
-      return if event_id.blank?
+      return if bib_prefix.present? || !bib_required? || event_id.blank?
 
-      self.bib_prefix = line_item.variant.bib_prefix
+      self.class.transaction do
+        rows = self.class.where(event_id: event_id, bib_prefix: line_item.variant.bib_prefix).lock
+        last_bib_number = rows.map(&:bib_number).max || 0
 
-      last_bib_number = event.guests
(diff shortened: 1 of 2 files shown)


hengsokly:
From bong Channa, we prefer to use the `optimistic lock`, but here we implement a pessimistic lock. So the solution is rejected.

PR #2183, opened 2024-12-24
Close #2180: enqueue generate BIB when order completed

### LINKED ISSUE #2180: Enqueue generate BIB

hengsokly:
- When payment is `completed`, the order will update the state to `complete` and trigger guest to `generate_bib!` (`SpreeCmCommissioner::OrderBibNumberConcern`)
- The bib_index is validated as unique
- Users may make payments simultaneously that make the guest_bib_index duplicate and raise errors.
- So applying to enqueue, but it may cause a delay and need to update the UI to handle the delay.

- **So need to verify with the cm-market-app to see how it affects the UI**


app/jobs/spree_cm_commissioner/order_complete_bib_generator_job.rb
```diff
@@ -0,0 +1,8 @@
+module SpreeCmCommissioner
+  class OrderCompleteBibGeneratorJob < ApplicationJob
+    def perform(order_id)
+      order = ::Spree::Order.find(order_id)
```
channainfo, line 4: Let's create a service to handle this.

channainfo (changes requested):

PR #2203, opened 2024-12-30
close #2202 improve set up google wallet

### LINKED ISSUE #2202: improve set up google wallet

nangdypanhar:
| Before | After |
|--------|-------|
| image| image|

[Screencast from 2025-01-03 10-02-05.webm](https://github.com/user-attachments/assets/cdf20c86-41fd-4783-9070-2a9f9150fb5f)

In this task :
. we enhance the set up with google wallet
. we set default for the input fields which help set up more faster 
. Also, we use only one button to verify with google

### DIFF
--- a/app/helpers/spree/admin/google_wallets_helper.rb
+++ b/app/helpers/spree/admin/google_wallets_helper.rb
@@ -1,0 +1,73 @@
+module Spree
+  module Admin
+    module GoogleWalletsHelper
+      # override
+      def preference_fields(object, form)
+        return unless object.respond_to?(:preferences)
+
+        fields = object.preferences.keys.map do |key|
+          next unless object.has_preference?(key)
+          next if %i[response verified_at].include?(key)
+
+          render_preference_field(object, form, key)
+        end
+        safe_join(fields)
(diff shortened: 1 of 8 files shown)


app/helpers/spree/admin/google_wallets_helper.rb
```diff
@@ -0,0 +1,55 @@
...
+        return unless object.respond_to?(:preferences)
+
+        fields = object.preferences.keys.map do |key|
+          next unless object.has_preference?(key)
+
+          next if key == :response
+          next if key == :verified_at
+
+          if key == :background_color
+            content_tag(:div, class: 'form-group', id: [object.class.to_s.parameterize, 'preference', key].join('-')) do
```
channainfo, line 15: extract the blocks inside the if/else to other methods.

channainfo (changes requested):

app/helpers/spree/admin/google_wallets_helper.rb
nangdypanhar: Resolve bong

PR #2205, opened 2025-01-02
close #2204 check item availability before add to queue

### LINKED ISSUE #2204: check item availability before add to queue

panhachom:
https://github.com/user-attachments/assets/b60c0f43-3154-40f4-934b-d36abb083cec

### DIFF
--- a/app/controllers/spree/api/v2/storefront/queue_cart/line_items_controller.rb
+++ b/app/controllers/spree/api/v2/storefront/queue_cart/line_items_controller.rb
@@ -13,6 +13,12 @@ module Spree
               spree_authorize! :update, spree_current_order, order_token
               spree_authorize! :show, @variant
 
+              availability_checker = SpreeCmCommissioner::Stock::AvailabilityChecker.new(@variant)
+
+              unless availability_checker.can_supply?(add_item_params[:quantity].to_i, add_item_params[:options])
+                return render_error_payload(availability_checker.error_message || I18n.t('variant_availability.items_out_of_stock'))
+              end
+
               job = SpreeCmCommissioner::EnqueueCart::AddItemJob.perform_later(
                 spree_current_order.id,
                 @variant.id,
--- a/app/queries/spree_cm_commissioner/variant_availability/non_permanent_stock_query.rb
+++ b/app/queries/spree_cm_commissioner/variant_availability/non_permanent_stock_query.rb
@@ -1,11 +1,12 @@
 module SpreeCmCommissioner
   module VariantAvailability
     class NonPermanentStockQuery
-      attr_reader :variant, :except_line_item_id
+      attr_reader :variant, :except_line_item_id, :error_message
 
       def initialize(variant:, except_line_item_id: nil)
         @variant = variant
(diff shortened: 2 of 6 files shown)


channainfo:
@panhachom, it would be helpful to show whether an item is out of stock or running low. Can we display the remaining quantity so users can make better purchasing decisions?

For example, if a user wants to buy 3 items but only 2 are left in stock, showing the remaining quantity would allow them to adjust their order and proceed without confusion. This would greatly improve the user experience.

panhachom:
> @panhachom, it would be helpful to show whether an item is out of stock or running low. Can we display the remaining quantity so users can make better purchasing decisions?
> 
> For example, if a user wants to buy 3 items but only 2 are left in stock, showing the remaining quantity would allow them to adjust their order and proceed without confusion. This would greatly improve the user experience.

resolve bong

app/queries/spree_cm_commissioner/variant_availability/non_permanent_stock_query.rb
```diff
@@ -20,10 +21,21 @@ def available?(quantity)
...
+        available_quantity = if reserve_variants.any?
+                               reserve_variants.sum(&:available_quantity)
+                             else
+                               variant.stock_items.sum(:count_on_hand)
+                             end
+
+        if available_quantity >= quantity
+          true
+        elsif available_quantity <= 0
+          @error_message = 'Item out of stock'
```
channainfo, line 34: use the locale string

app/queries/spree_cm_commissioner/variant_availability/non_permanent_stock_query.rb
```diff
@@ -20,10 +21,21 @@ def available?(quantity)
...
+          false
         else
-          variant.stock_items.sum(:count_on_hand) >= quantity
+          @error_message = "Only #{available_quantity} item(s) available"
```
channainfo, line 37: use locale string instead.

```
 @error_message = I18n.t('variant_availability.item_available_instock') :  I18n.t('variant_availability.items_available_instock', available_quantity: available_quantity) 
```
we should have two keys under variant_availability


```
variant_availability:
  item_available_instock: Only 1 item available in stock.
  items_available_instock:  Only %{available_quantity} items available in stock.
   
```

channainfo (changes requested):

app/controllers/spree/api/v2/storefront/queue_cart/line_items_controller.rb
```diff
@@ -13,6 +13,12 @@ def create
               spree_authorize! :update, spree_current_order, order_token
               spree_authorize! :show, @variant
 
+              availability_checker = SpreeCmCommissioner::Stock::AvailabilityChecker.new(@variant)
+
+              unless availability_checker.can_supply?(add_item_params[:quantity].to_i, add_item_params[:options])
+                return render_error_payload(availability_checker.error_message || 'Item out of Stock')
```
channainfo, line 19: Move this string "Item out of Stock" to locale as well

channainfo (changes requested):

PR #2211, opened 2025-01-02
close #2176: add API to support guest account and transfer cart

### LINKED ISSUE #2176: Support Guest checkout
Allow guest user to add items to cart and then transfer the item once the user decides to log in

Gem:  https://github.com/cbeer/devise-guests.

hengsokly:
**Support Adding Items to Cart for Guest Users**

- **API to Create Guest Users**: Allows the creation of temporary guest user accounts for interacting with the cart.
- **API for Cart Transfer**: Facilitates transferring items in the cart from a guest user to a logged-in user upon authentication.


app/controllers/spree/api/v2/storefront/cart_transfers_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    module V2
+      module Storefront
+        class CartTransfersController < ::Spree::Api::V2::ResourceController
+          include OrderConcern
+
+          before_action :require_spree_current_user
+          before_action :load_guest
+
+          def update
+            if spree_current_order&.update(user: spree_current_user)
```
channainfo, line 12: Let's move all the logic to a service or an interactor called GuestCartTransfer given the guest_user and the current user, and the sequence of action might look like this.

- Get the cart of the guest
- Transfer the cart to the current user
- Clean Guest data ( access_token )
- Post transfer cart action ( notification, write log ).

app/controllers/spree/api/v2/storefront/cart_transfers_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+              store: current_store,
+              user: @guest_user,
+              user_id: @guest_user.try(:id),
+              token: order_token,
+              currency: current_currency
+            )
+          end
+
+          def load_guest
+            @guest_user = Spree::User.find_by(secure_token: params[:guest_token], guest: true)
```
channainfo, line 33: To improve security we can take 2 approaches:

1. the guest user signs the request with JWT token using guest_token
2. send a more field to ( guest_user.id, guest_user.guest_token). Find the guest by id and then match if the guest_token is valid.

let's always try to find the user by id first ( id is always index ) and then comparing with with other field ( secure_token).

The 1 approach is the most secure, and the 2 approach is easy to implement.

app/models/concerns/spree_cm_commissioner/user_guest.rb
```diff
@@ -0,0 +1,15 @@
+module SpreeCmCommissioner
+  module UserGuest
+    extend ActiveSupport::Concern
+
+    included do
+      def self.create_guest
+        create(email: "guest_#{SecureRandom.hex(10)}@example.com", guest: true, confirmed_at: Time.zone.now, secure_token: SecureRandom.uuid)
```
channainfo, line 7: let's increase the entropy of the email SecureRandom.hex(24).

app/models/concerns/spree_cm_commissioner/user_guest.rb
```diff
@@ -0,0 +1,15 @@
...
+      end
+
+      def password_required?
+        super && !guest?
```
channainfo, line 11: This is dangerous if we don't test it well and the behaviour is not clear. Please check the gem how to handle this or we can just generate a secure random password 12 chars instead.

spec/request/spree/api/v2/controller/account_guests_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    it "returns status 201" do
+      post :create
+      expect(response.status).to eq 201
+    end
+
+    it "returns a guest access_token" do
+      post :create
+      expect(JSON.parse(response.body)["access_token"]).not_to be_nil
```
channainfo, line 16: validate newly created field as well.

spec/request/spree/api/v2/controller/account_guests_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe Spree::Api::V2::Storefront::AccountGuestsController, type: :controller do
```
channainfo, line 3: To effectively test an API, using **request specs** is preferable because:

- They are designed for **integration testing**, ensuring end-to-end functionality.  
- **Controller specs** do not handle routing, making them incomplete for API testing. Even if a controller spec passes, it doesn’t guarantee the API works correctly, as route misconfigurations might still exist.

channainfo (changes requested):

app/interactors/spree_cm_commissioner/guest_cart_transfer.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+    end
+
+    private
+
+    def transfer_guest_cart_to_user!
+      if user_cart.present?
+        merge_cart!
+      else
+        guest_cart.update!(user: user)
```
channainfo, line 24: Let's move this block to a separate method called replace_cart!

app/interactors/spree_cm_commissioner/guest_cart_transfer.rb
```diff
@@ -0,0 +1,78 @@
...
+      rescue StandardError => e
+        context.fail!(message: e.message)
+        raise ActiveRecord::Rollback
+      end
+    end
+
+    private
+
+    def transfer_guest_cart_to_user!
+      if user_cart.present?
```
channainfo, line 21: It seems merging cart is complicated. Let's add and option to customize the behavior by merging or replacing the cart to mitigate outages.

channainfo (changes requested):

app/interactors/spree_cm_commissioner/guest_cart_transfer.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def user_cart
+      @user_cart ||= find_current_order(user.id)
+    end
+
+    def find_current_order(user_id)
+      Spree::Api::Dependencies.storefront_current_order_finder.constantize.new.execute(
+        store: store,
+        user: guest,
+        user_id: user_id,
```
hengsokly, line 65: Double check with spree code

app/interactors/spree_cm_commissioner/guest_cart_transfer.rb
```diff
@@ -0,0 +1,78 @@
...
+        # destroyed, we need to reload the guest record:
+        guest.reload
+      end
+    end
+
+    def merge_cart!
+      user_items = user_cart.line_items.index_by(&:variant_id)
+
+      guest_cart.line_items.reload.each do |guest_item|
+        existing_item = user_items[guest_item.variant_id]
```
hengsokly, line 36: Refactor login in loop in to a small method, and write spec

spec/models/concerns/spree_cm_commissioner/guest_user_concern_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'spec_helper'
+
+RSpec.describe 'SpreeCmCommissioner::GuestUserConcern' do
```
hengsokly, line 3: will move test to sprec JWT class above

spec/models/concerns/spree_cm_commissioner/guest_user_concern_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  describe '.create_guest' do
+    it 'create a guest user' do
+      expect { Spree::User.create_guest }.to change {
+        Spree::User.where(guest: true).count
+      }.by(1)
+    end
+  end
```
hengsokly, line 10: Remove the concern spec, and move this one into user spec

app/controllers/spree/api/v2/storefront/guest_cart_transfers_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    module V2
+      module Storefront
+        class GuestCartTransfersController < ::Spree::Api::V2::ResourceController
+          def update
+            result = SpreeCmCommissioner::GuestCartTransfer.call(guest_cart_params)
+            if result.success?
+              head :ok
+            else
+              render json: { error: result.message }, status: :unprocessable_entity
+            end
```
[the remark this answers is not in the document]
hengsokly: - Call Iterator ClassName and its params: Done
- Move methods to a base class: Base on time, this code is acceptable and can be improved later.

app/interactors/spree_cm_commissioner/guest_cart_transfer.rb
```diff
@@ -0,0 +1,128 @@
...
+    end
+
+    private
+
+    def validate_merge_type!
+      context.fail!(error: "Invalid merge type: #{merge_type}") unless MERGE_TYPES.values.include?(merge_type)
+    end
+
+    def validate_guest_presence!
+      context.fail!(error: 'Guest is invalid!') if guest.nil?
```
channainfo, line 36: Let's verify the consistency with other interactors. I think we should use **message:** instead of **error:** like this:

```
context.fail!(message: error_message)
```

Maintaining the same key is crucial to ensure consistency throughout the codebase and prevent any potential naming conflicts.

channainfo (changes requested):

PR #2222, opened 2025-01-06
close#2168 display map in venue set up

nangdypanhar:


channainfo:
Set the default map zoom level to 18 and enable the pushpin marker to be draggable. When the marker is moved, the latitude and longitude values should automatically update in the input fields.

Additionally, any changes made to the latitude and longitude inputs should reflect on the marker's position, ensuring two-way synchronization between the marker and the input fields.

PR #2230, opened 2025-01-09
close #2229 missing translation on link account fail

### LINKED ISSUE #2229: Missing translation on link account fail

panhachom:
Image


channainfo:
@panhachom let's use i18n-tasks gem to help sort this out.

PR #2244, opened 2025-01-10
close #2236 fix noticed response content cachable

### LINKED ISSUE #2236: Noticed::ResponseUnsuccessful: Noticed::ResponseUnsuccessful (Noticed::ResponseUnsuccessful)
Sentry Issue: [RUBY-RAILS-1W](https://bookmeplus.sentry.io/issues/6207341527/?referrer=github_integration)

```
Noticed::ResponseUnsuccessful: Noticed::ResponseUnsuccessful (Noticed::ResponseUnsuccessful)
  noticed (1.6.3) lib/noticed/delivery_methods/base.rb:88:in `post'
  noticed (1.6.3) lib/noticed/delivery_methods/fcm.rb:21:in `block in deliver'
  noticed (1.6.3) lib/noticed/delivery_methods/fcm.rb:20:in `each'
  noticed (1.6.3) lib/noticed/delivery_methods/fcm.rb:20:in `deliver'
  noticed (1.6.3) lib/noticed/delivery_methods/base.rb:55:in `block in perform'
...
(160 additional frame(s) were not displayed)
```

LengTech11:
## This PR is 
Explanation:
This method sets cache control headers for the response.
- `return if response.committed?`: Exits the method if the response is already committed (that is headers have already been sent).
-` return if response.status != 200`: Exits the method if the response status is not 200 (OK).
-  If the response is not committed and the `status is 200`, it sets the following headers:
- `Cache-Control`: Specifies caching directives. public makes the response cacheable by any cache, max-age=0 means the browser cache should not cache the response, and `s-maxage=#{max_age}` sets the server cache duration.
`Pragma`: Sets to no-cache for older HTTP/1.0 clients to prevent caching.
`Expires`: Sets to 0 to indicate that the response is already expired.

## DEMO Postman Headers

https://github.com/user-attachments/assets/25f34804-89ce-4db1-b28f-a007ea7a3c8a

### DIFF
--- a/app/controllers/concerns/spree_cm_commissioner/content_cachable.rb
+++ b/app/controllers/concerns/spree_cm_commissioner/content_cachable.rb
@@ -11,6 +11,9 @@ module SpreeCmCommissioner
     end
 
     def set_cache_control_for_cdn
+      return if response.committed?
+      return if response.status != 200
+
       # max-age: browser cache, s-maxage: server cache
       response.headers['Cache-Control'] = "public, max-age=0, s-maxage=#{max_age}"
       response.headers['Pragma'] = 'no-cache' # For older HTTP/1.0 clients


app/controllers/concerns/spree_cm_commissioner/content_cachable.rb
```diff
@@ -3,14 +3,17 @@ module ContentCachable
...
+      skip_before_action :set_cache_control_for_cdn, only: %i[stream_file proxy]
     end
 
     def max_age
       ENV.fetch('CONTENT_CACHE_MAX_AGE', '180')
     end
 
     def set_cache_control_for_cdn
+      return if response.committed?
+
```
channainfo, line 16: let's add another one
```
   return if response.status == 200
```

app/controllers/concerns/spree_cm_commissioner/content_cachable.rb
```diff
@@ -3,14 +3,17 @@ module ContentCachable
     extend ActiveSupport::Concern
 
     included do
-      after_action :set_cache_control_for_cdn
+      before_action :set_cache_control_for_cdn
```
channainfo, line 6: let's keep after_action so rollback this block

channainfo (changes requested):

channainfo:
@LengTech11 you've got a good summary up there. Nice!

LengTech11:
## DEMO Fail to save Cache

https://github.com/user-attachments/assets/514ab486-3c9b-4d04-85fb-033742d3cc7d

PR #2250, opened 2025-01-10
Close #2249 Add section tab in taxon edit screen

### LINKED ISSUE #2249: Add section tab in taxon edit screen
with this section admin can set up taxon and product easier and faster.

mengthong-ly:
Frame 1

### DIFF
--- a/app/controllers/spree/admin/taxon_childrens_controller.rb
+++ b/app/controllers/spree/admin/taxon_childrens_controller.rb
@@ -1,0 +1,20 @@
+module Spree
+  module Admin
+    class TaxonChildrensController < Spree::Admin::ResourceController
+      before_action :load_taxonomy_taxon, only: [:index]
+
+      def index; end
+
+      private
+
+      def model_class
+        Spree::Taxon
+      end
+
+      def load_taxonomy_taxon
+        @taxonomy = Spree::Taxonomy.find(params[:taxonomy_id])
+        @taxon = @taxonomy.taxons.find(params[:taxon_id])
+      end
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/controllers/spree/admin/view_events_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Spree
+  module Admin
+    class ViewEventsController < Spree::Admin::ResourceController
+      before_action :load_taxonomy_taxon
```
channainfo, line 4: explicitly add which action to apply:
```
before_action :load_taxonomy_taxon, only: [:index]
```

app/controllers/spree/admin/view_events_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      def index
+        # Fetch the specific taxon associated with the taxonomy
+        @taxon = @taxonomy.taxons.find(params[:taxon_id])
```
channainfo, line 8: you don't need this because the load_taxonomy_taxon already does just this.

app/models/spree_cm_commissioner/view_event.rb
```diff
@@ -0,0 +1,4 @@
+module SpreeCmCommissioner
+  class ViewEvent < ApplicationRecord
```
channainfo, line 2: remove this, the model is linked to a table in the DB.

app/controllers/spree/admin/view_events_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Spree
+  module Admin
+    class ViewEventsController < Spree::Admin::ResourceController
```
channainfo, line 3: Why not decorating the existing TaxonController? Naming a new controller can be challenging, and if the new controller is minimal, the additional effort might not be justified.

channainfo (changes requested):

PR #2253, opened 2025-01-11
close #2252 remove passed events from homepage section in app

### LINKED ISSUE #2252: remove passed events from homepage section app

nangdypanhar:
| Before | After |
|--------|-------|
| image | image|

[Screencast from 2025-01-21 10-02-00.webm](https://github.com/user-attachments/assets/e65fc817-cd5e-4fec-a760-14a37405c1b2)

Description :
- In this PR , we create an automatic event  available_on  and discontinue_on base on the date that the user input 
- The event will automatic show and hide on the homepage section base on the date that have been set,
- if the user that does not set available_on and discontinue_on dates ,  the event will always active on the homepage section

### DIFF
--- a/app/controllers/spree/admin/base_controller_decorator.rb
+++ b/app/controllers/spree/admin/base_controller_decorator.rb
@@ -34,23 +34,34 @@ module Spree
       # POST
       def invalidate_api_caches
         if params[:model].present?
-          api_patterns_map = {
-            'SpreeCmCommissioner::HomepageSection' => '/api/v2/storefront/homepage/*',
-            'SpreeCmCommissioner::HomepageBackground' => '/api/v2/storefront/homepage/*',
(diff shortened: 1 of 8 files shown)


db/migrate/20250111100028_add_show_at_and_hide_at_to_homepage_section_relatables.rb
```diff
@@ -0,0 +1,6 @@
+class AddShowAtAndHideAtToHomepageSectionRelatables < ActiveRecord::Migration[7.0]
+  def change
+    add_column :cm_homepage_section_relatables, :show_at , :datetime , if_not_exists: true
```
channainfo, line 3: This seems redundant. Let's review how it's implemented in the product and align the field names accordingly.

channainfo (changes requested):

PR #2255, opened 2025-01-11
Close #2247: stop retry fcm queue when missing notification record

### LINKED ISSUE #2247: From ActiveJob::DeserializationError #2081, implement solution 1
From ActiveJob::DeserializationError #2081, implement solution 1 to reduce the queue retry traffic

**Rescue Missing Record:**
Modify the notification code to handle cases where the record is not found:
`rescue ActiveRecord::RecordNotFound`

hengsokly:
- Added `DeliveryMethods::FcmDecorator` to enhance the functionality of FCM delivery.
- Implemented a discard job mechanism to handle scenarios where a notification record is missing, ensuring that unnecessary processing is avoided.

### DIFF
--- a/app/jobs/spree_cm_commissioner/application_job_decorator.rb
+++ b/app/jobs/spree_cm_commissioner/application_job_decorator.rb
@@ -1,0 +1,18 @@
+module SpreeCmCommissioner
+  module ApplicationJobDecorator
+    def self.prepended(base)
+      base.discard_on ActiveJob::DeserializationError do |job, error|
+        handle_deserialization_error(job, error)
+      end
+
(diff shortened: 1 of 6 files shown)


app/notifications/noticed_fcm_base.rb
```diff
@@ -11,7 +11,6 @@ def format_for_database
       params: {
         payload: payload,
         translatable_options: translatable_options
```
channainfo, line 13: Check out this file, there is no change needed here.

channainfo (changes requested):

channainfo:
@hengsokly it looks great, just a tiny comment to checkout a file that has nothing to do with the issue.

PR #2259, opened 2025-01-13
close #2258 add index to phone_number spree_users

### LINKED ISSUE #2258: Enhance users
Here is one of the top query

```
SELECT ? AS one FROM "spree_users" 
WHERE "spree_users"."phone_number" = $1 AND "spree_users"."deleted_at" IS NULL AND "spree_users"."id" != $2 LIMIT $3

```

LengTech11:


channainfo:
@LengTech11 we should user intel phone number instead of phone number.

db/migrate/20250113043029_add_index_to_spree_user_phone_number.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexToSpreeUserPhoneNumber < ActiveRecord::Migration[7.0]
+  def change
+    add_index :spree_users, :phone_number, if_not_exists: true
```
channainfo, line 3: let's not index phone_number. use the intel_phone_number instead. we should query by intel_phone_number not the phone_number

channainfo (changes requested):

spec/models/spree/order_spec.rb
```diff
@@ -428,12 +428,12 @@
 
   describe '#associate_user' do
     let!(:order) { create(:order) }
-    let(:user_with_phone_number) { create(:user , phone_number: '012290564') }
```
channainfo, line 431: We pass the `phone_number` to the model, and in the `before_validation` callback, we use a library to derive the `intel_phone_number` from the `phone_number` and set it accordingly.

channainfo (changes requested):

PR #2262, opened 2025-01-13
close #2257 cache store url and enchance by_url scope

### LINKED ISSUE #2257: Enhance Spree store
currently, the following query is triggered ( before_action)

```
SELECT "spree_stores".*
FROM "spree_stores"
WHERE "spree_stores"."deleted_at" IS NULL AND (url like '%api-production.bookme.plus%')
ORDER BY "spree_stores"."created_at" ASC
LIMIT $1
```

let's optimize:
- return the **like** to use an  exact match
- or we can just return the first store.

and we can cache the store.

Using the exact match may not work if we use multiple domains ( images.bookme.plus, static.bookme.plus)

LengTech11:
### Description
The `current_store` method is an override that determines the current store based on the request's host. It first attempts to find a store with an exact match for the URL. If no exact match is found, it falls back to a partial match. If neither an exact nor a partial match is found, it returns the default store. The result is cached for 1 hour to avoid repeated database queries.

### Implementation
```ruby
def current_store
  @current_store ||= Rails.cache.fetch("current_store_#{request.host}", expires_in: 1.hour) do
    Spree::Store.where('url = ? OR url LIKE ?', request.host, "%#{request.host}%").first || Spree::Store.default
  end
end

### DIFF
--- a/app/controllers/spree/admin/base_controller_decorator.rb
+++ b/app/controllers/spree/admin/base_controller_decorator.rb
@@ -5,6 +5,15 @@ module Spree
         base.before_action :redirect_to_billing, if: -> { Spree::Store.default.code.include?('billing') }
       end
 
+      # override
+      # Default behavior is to return default store based on current URL.
+      # Our admin URL is not store URL, so we return default store instead.
+      def current_store
+        @current_store ||= Rails.cache.fetch("current_store_#{request.host}", expires_in: 1.hour) do
+          Spree::Store.where('url = ? OR url LIKE ?', request.host, "%#{request.host}%").first || Spree::Store.default
+        end
+      end
+
       # even db in read only mode, authorizor still need to run Warden callbacks which will write to db
       # this include update tracked fields such as sign_in_count, last_sign_in_at, etc.
       def authorize!(*args)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/models/spree_cm_commissioner/store_decorator.rb [hunk 1]
```diff
@@ -5,6 +5,22 @@ def self.prepended(base)
...
+
+      # Override by_url scope
+      # Optimized by_url scope with exact match, partial match, adn caching
+      # Cache the result for 1 hour to avoid repeated queries
+      # First attempt exact match, then partial match if no exact match is found
+      # Ensure we always return the default store when url is nil or empty
+      base.scope :by_url, lambda { |url|
+        Rails.cache.fetch("store_by_url_#{url}", expired_in: 1.hour) do
+          if url.present?
+            store = find_by('url = ?', url) || where('url like ?', "%#{url}%").first
```
channainfo, line 17: Just use one statement.
```
where('url = ? OR url like ?', url, "%#{url}%").first
```

channainfo (changes requested):

channainfo:
No longer needed.

PR #2264, opened 2025-01-13
close #2263 update stock items in real time

panhachom:


panhachom:
> @panhachom let's deploy this to staging and test the app and the admin panel.

yeah bong i will

PR #2313, opened 2025-01-31
Close #2284: Pre-calculated user cart_item_count and unread_notificat…

### LINKED ISSUE #2284: Enhance app booting with precalculate
- The number of items in the cart.
- The number of unread notifications.

hengsokly:
…ion_count

**The scenario is as following**
- We want a user to have the number of items in the cart and the number of unread notifications.
- When user `add/remove line_item` from `current cart`, we update the user `cart_item_count`
- When user complete the order, we reset the user `cart_item_count`
- When `create/mark_as_read` notification, we update the user `unread_notification_count`


app/interactors/spree_cm_commissioner/notification_base_reader.rb
```diff
@@ -0,0 +1,33 @@
+module SpreeCmCommissioner
+  class NotificationBaseReader < BaseInteractor
+    delegate :id, :user, to: :context
+
+    def call
+      validation!
+
+      ActiveRecord::Base.transaction do
+        mark_as_read!
+        decrement_user_unread_notification_count!
```
channainfo, line 10: @hengsokly let's create another method called decrement_unread_count!
```
def decrement_unread_count!
  decrement_user_unread_notification_count!
  consolidate_unread_count_async
end
```

app/jobs/spree_cm_commissioner/user_unread_notification_count_job.rb
```diff
@@ -0,0 +1,11 @@
+module SpreeCmCommissioner
+  class UserUnreadNotificationCountJob < ApplicationUniqueJob
+    queue_as :user_updates
+
+    def perform(user_id)
+      user = Spree::User.find_by(id: user_id)
```
channainfo, line 6: user could be nil because in our system users can request to delete their account. let's handle the nil scenario.

app/models/spree_cm_commissioner/order_updater_decorator.rb
```diff
@@ -1,10 +1,24 @@
...
+    def update
+      super
+
+      # After the order is updated including item_count column
+      update_user_cart_item_count
+    end
+
     # override
     def update_item_total
       order.item_total = line_items.sum(&:amount)
```
channainfo, line 13: why we need to override this method?

app/models/spree_cm_commissioner/order_updater_decorator.rb
```diff
@@ -1,10 +1,24 @@
...
       update_order_total
     end
+
+    private
+
+    def update_user_cart_item_count
+      SpreeCmCommissioner::UserCartItemCountUpdater.new.call(order: order)
```
channainfo, line 20: what if the order state does not change? the cart item count should  get updated if it is a cart. I think when a item is added or remove from the cart or the payment has been made then the cart item count should get updated.

- The last non complete order is a cart.

app/interactors/spree_cm_commissioner/update_user_cart_item_count_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+    def reset_user_cart_item_count!
+      user.update(cart_item_count: 0)
+    end
+
+    def current_cart?
+      order == user_current_cart
+    end
+
+    def user_current_cart
+      @user_current_cart ||= Spree::Api::Dependencies.storefront_current_order_finder.constantize.new.execute(
```
channainfo, line 57: if we can get the user current cart, it seems we don't need the order in the parameter. what do you think?

channainfo (changes requested):

PR #2327, opened 2025-02-05
close #2323 Tenant SignUp with PinCode and Login with Email and Password

LengTech11:
This PR includes the following changes:
- Add Foreign Key to SpreeUser Tenant_id
- Scoped User data (e.g., login, phone_number) by Tenant for uniqueness.
- Added RSpec tests for users create with pin_code with the same login or phone_number but under different tenants.
- Created new APIs for tenant-specific actions, such as:
     - user_registration_with_pin_codes
     - account_checkers
     - pin_code_generators
- Reused existing Interactor logic across the new API endpoints.

### DEMO

https://github.com/user-attachments/assets/3378a667-3b60-455a-a79e-cc89837f2c05


app/models/spree_cm_commissioner/user_decorator.rb
```diff
@@ -26,6 +26,10 @@ def self.prepended(base)
 
       base.has_secure_password :confirm_pin_code, validations: false
 
+      base.multi_tenant :tenant, class_name: 'SpreeCmCommissioner::Tenant'
+      # Define the scope to filter users by tenant
```
theachoem, line 30: no need comment here. code should describe itself already.

app/controllers/spree/api/v2/tenant/account_checker_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Spree
+  module Api
+    module V2
+      module Tenant
+        class AccountCheckerController < BaseController
+          def index
+            checker = SpreeCmCommissioner::ExistingAccountChecker.call(filter_params.to_h)
```
theachoem, line 7: maybe we don't need this API, let remove it if not used. Keep it simple.

app/controllers/spree/api/v2/tenant/pin_code_checkers_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Spree
+  module Api
+    module V2
+      module Tenant
+        class PinCodeCheckersController < BaseController
+          def update
+            context = SpreeCmCommissioner::PinCodeChecker.call(pin_code_attrs)
```
theachoem, line 7: this is also not used. let's remove it.

app/services/spree_cm_commissioner/user_authenticator.rb
```diff
@@ -7,7 +7,15 @@ def self.call!(params)
       context = auth_context(params)
       raise exception(context.message) unless context.success?
 
-      context.user
+      oauth_application = Spree::OauthApplication.find_by(uid: params[:client_id],
```
theachoem, line 10: Let's not pass client id or secret to here, find way to pass oauth application from doorkeeper directly, make it an optional params.

config/routes.rb
```diff
@@ -398,6 +398,10 @@
 
       namespace :tenant do
         resources :vendors
+        resources :account_checker
+        resource :pin_code_checkers, only: [:update]
+        resource :user_registration_with_pin_codes, only: [:create]
+        resources :pin_code_generators, only: [:create]
```
theachoem, line 404: Just this 2 API should work for  registration:
- user_registration_with_pin_codes
- pin_code_generators

theachoem (changes requested):

config/routes.rb
LengTech11: we also use pin code checkers base on our App Ui

PR #2329, opened 2025-02-06
Close #2328 fix email sso add email to user

### LINKED ISSUE #2328: Fix Gmail SSO add email to user

SreyMochOrng:


app/interactors/spree_cm_commissioner/user_identity_checker.rb
```diff
@@ -16,6 +16,11 @@ def load_user
         sub: context.sub
       ).first
 
+      if user_identity_provider.present? && user_identity_provider.user.email.nil? && user_identity_provider.email.present?
+        user_identity_provider.user.email = user_identity_provider.email
+        user_identity_provider.user.save
```
theachoem, line 21:
```rb
user_identity_provider.user.email ||= user_identity_provider.email
user_identity_provider.user.save if user_identity_provider.user.email_changed?
```

theachoem: it is not recommend to save in checker like this. it is suppose to be read only. can you put in the authenticator instead?

theachoem (changes requested):

app/interactors/spree_cm_commissioner/user_identity_checker.rb
channainfo: @SreyMochOrng when there is a bug, a spec to prove the fix as well

PR #2333, opened 2025-02-07
close # 2332 handle insufficient stock on order resume

nangdypanhar:


app/controllers/spree_cm_commissioner/admin/orders_controller_decorator.rb
```diff
@@ -79,6 +79,17 @@ def update_order_status
         redirect_back fallback_location: spree.edit_admin_order_url(@order)
       end
 
+      # override
+      def resume
+        if @order.line_items.any? { |item| item.quantity > item.variant.total_on_hand }
+          flash[:error] = I18n.t('orders.resume.insufficient_stock')
+        else
+          @order.resume!
```
theachoem, line 87: how about 

```rb
resumed = order.resume

if (resumed) {
  flash[:success] = ...
} else {
  flash[:error] = ...
}
```

config/locales/en.yml
```diff
@@ -154,6 +154,9 @@ en:
       exception_rules: "Exception Rules"
 
   orders:
+    resume:
```
theachoem, line 157: maybe no need, order.errors should return it.

theachoem (changes requested):

PR #2362, opened 2025-02-17
close #2360 Tenant Scope Overrided

### LINKED ISSUE #2360: Tenant Scope Override

LengTech11:
- I've overridden the scope to raise an error. and Alway let this child Override scope
Screenshot 2025-02-18 at 2 33 18 in the afternoon

### DIFF
--- a/app/controllers/spree/api/v2/tenant/base_controller.rb
+++ b/app/controllers/spree/api/v2/tenant/base_controller.rb
@@ -28,6 +28,11 @@ module Spree
               yield
             end
           end
+
+          # override
+          def scope
+            raise 'scope should be implemented in a sub-class'
+          end
         end
       end
     end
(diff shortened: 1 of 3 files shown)


app/controllers/spree/api/v2/tenant/base_controller.rb
```diff
@@ -28,6 +28,12 @@ def set_tenant_context
               yield
             end
           end
+
+          # override
+          # Let the child controller override the `scope` method to define the query
+          def scope
+            model_class.where(tenant_id: MultiTenant.current_tenant_id)
```
channainfo, line 35: let's raise exception instead. If we don't raise the exception, the child controller might forget to override the scope.

channainfo (changes requested):

app/controllers/spree/api/v2/tenant/base_controller.rb
LengTech11: 1 more solution i think we just override the SpreeResourceController Scope and Let Child Scope can be optional to define.

PR #2372, opened 2025-02-20
Close #2369 reserved stock method

### LINKED ISSUE #2369: Create reserved_stock and total_stock method

SSambo12:
In this PR, we create total_reserved_stock method to count total reserved stock of a product.

### DIFF
--- a/app/models/spree_cm_commissioner/product_decorator.rb
+++ b/app/models/spree_cm_commissioner/product_decorator.rb
@@ -75,6 +75,10 @@ module SpreeCmCommissioner
       taxons.event.first&.parent
     end
 
+    def total_reserved_stock
+      line_items.complete.where(variant_id: variants.pluck(:id)).sum(:quantity)
+    end
+
     private
 
     def set_tenant


app/models/spree_cm_commissioner/product_decorator.rb
```diff
@@ -75,6 +75,10 @@ def associated_event
       taxons.event.first&.parent
     end
 
+    def total_reserved_stock
+      Spree::LineItem.complete.where(variant_id: variants.pluck(:id)).sum(:quantity)
```
channainfo, line 79: let's not do this this  here. If you need this for stats, create a new class to return this.

channainfo (changes requested):

app/models/spree_cm_commissioner/product_decorator.rb
SSambo12: Hello bong @channainfo, I create this method to call that function in organizer dashboard, on ticket table section. And we use that method to count total reserved stock of each product.

image

PR #2374, opened 2025-02-20
close #2371 Remove email Unique index Constraint from User And Scope it By Tenant

### LINKED ISSUE #2371: Remove email Unique index Constraint from User And Scope it By Tenant

LengTech11:
- [x] Remove Email Unique constraint for spree_users
- [x] Handle Uniqueness scope Email with Tenant

Result:
Screenshot 2025-02-20 at 1 09 42 in the afternoon

### DIFF
--- a/app/models/concerns/spree_cm_commissioner/user_identity.rb
+++ b/app/models/concerns/spree_cm_commissioner/user_identity.rb
@@ -15,6 +15,13 @@ module SpreeCmCommissioner
 
       validates :login, presence: true, if: -> (u) { u.require_login_identity_all_blank_for(:login) }
       validates :login, uniqueness: { scope: :tenant_id, case_sensitive: false, allow_blank: true }
+
+      _validators.reject! { |key, _| key == :email }
+      _validate_callbacks.each { |c| c.filter.attributes.delete(:email) if c.filter.respond_to?(:attributes) }
+
+      validates :email, presence: true, if: -> (u) { u.require_login_identity_all_blank_for(:email) }
+      validates :email, uniqueness: { scope: :tenant_id, case_sensitive: false, allow_blank: true }
+      validates :email, format: { with: URI::MailTo::EMAIL_REGEXP, allow_blank: true }
     end
 
     class_methods do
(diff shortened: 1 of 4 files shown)


app/models/spree_cm_commissioner/user_decorator.rb
```diff
@@ -120,5 +120,12 @@ def update_otp_enabled
   end
 end
 
-Spree::User.prepend(SpreeCmCommissioner::UserDecorator) unless Spree::User.included_modules.include?(SpreeCmCommissioner::UserDecorator)
+if Spree::User.included_modules.exclude?(SpreeCmCommissioner::UserDecorator)
+  # remove all email validations
```
channainfo, line 124: Put this block in the User decorator
```
_validators.reject! { |key, _| key == :email }
_validate_callbacks.each { |c| c.filter.attributes.delete(:email) if c.filter.respond_to?(:attributes) }
```

to avoid fragmenting code in multiple places

channainfo (changes requested):

app/models/spree_cm_commissioner/user_decorator.rb
LengTech11: Hello bong @channainfo this Block code already include this code bong.

- user_decorator.rb
```rb
if Spree::User.included_modules.exclude?(SpreeCmCommissioner::UserDecorator)
  # remove all email validations
  Spree::User._validators.reject! { |key, _| key == :email }
  Spree::User._validate_callbacks.each { |c| c.filter.attributes.delete(:email) if c.filter.respond_to?(:attributes) }

  # prepend decorator will include new validations
  Spree::User.prepend(SpreeCmCommissioner::UserDecorator)
end
# rubocop:enable Metrics/MethodLength
```

app/models/spree_cm_commissioner/user_decorator.rb
```diff
@@ -1,8 +1,12 @@
 # rubocop:disable Metrics/MethodLength
 module SpreeCmCommissioner
   module UserDecorator
-    def self.prepended(base)
+    def self.prepended(base) # rubocop:disable Metrics/AbcSize
       base.include SpreeCmCommissioner::UserNotification
+
+      base._validators.reject! { |key, _| key == :email }
```
channainfo, line 7: move this block to UserIdentity module instead because they are correlated with removing and adding new logic.

channainfo (changes requested):

PR #2586, opened 2025-04-29
Close #2584 implement CreateEvent service

SSambo12:


app/interactors/spree_cm_commissioner/create_event.rb
```diff
@@ -0,0 +1,63 @@
+module SpreeCmCommissioner
+  class CreateEvent
+    include Interactor
+    delegate :params, :current_vendor, to: :context
+
+    def call
+      build_parent_taxon
```
channainfo, line 7: Wrap it within the transaction ensure the data integrity. context.fail when the transaction is not successfully committed.

channainfo (changes requested):

PR #2625, opened 2025-05-09
Close #2624 implement create ticket service

### LINKED ISSUE #2624: Implement CreateTicket service

SSambo12:
This PR, we implement a `CreateTicket` service that handles the ticket creation process in one transaction. The process include: 

- Create the ticket

- Link ticket to the event

- Create variant

- Add stock

If any step fails, the whole process is rolled back to prevent data from being saved.

### DIFF
--- a/app/interactors/spree_cm_commissioner/create_ticket.rb
+++ b/app/interactors/spree_cm_commissioner/create_ticket.rb
@@ -1,0 +1,95 @@
+module SpreeCmCommissioner
+  class CreateTicket
+    include Interactor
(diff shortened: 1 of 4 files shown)


app/interactors/spree_cm_commissioner/create_ticket.rb
```diff
@@ -0,0 +1,93 @@
+module SpreeCmCommissioner
```
channainfo, line 1: Write a spec for it

channainfo (changes requested):

PR #2641, opened 2025-05-13
Close #2638 migrate to not store checkout date

### LINKED ISSUE #2638: Migrate to not storing checkout date in line item.

theachoem:
## Problem
Previously, to actual from_date or to_date of line_item, we do conditioning as following:
- For accommodation: from_date...to_date - 1 day (exclude the checkout date)
- For other product type: from_date...to_date

This looks good at the start, but in code base, we will see us checking this condition everywhere just to date the actual date. 

## Solution
Store actual date in from_date & to_date
- For accommodation: from_date & to_date are stay date
- For transit: from_date & to_date are the same date
- For service: from_date & to_date are service date

So purpose from_date & to_date is now unify and easy to use, no need to check condition everywhere.

## Next
- We need to search hotel by stay date instead of check-in & checkout date
- We also need to pass stay date to create line item instead of checkin/out date
- After purchase, we also have to display :checkout_date instead of :to_date.

### DIFF
--- a/app/finders/spree_cm_commissioner/accommodations/find.rb
+++ b/app/finders/spree_cm_commissioner/accommodations/find.rb
@@ -13,7 +13,7 @@ module SpreeCmCommissioner
       def execute
         scope
           .where(default_state_id: state_id)
-          .where(inventory_items: { inventory_date: date_range_excluding_checkout })
+          .where(inventory_items: { inventory_date: stay_dates })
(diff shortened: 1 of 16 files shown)


theachoem:
About 49 orders have been confirmed since 2024. So we don't have to do migration for old data. We can manually edit if needed.

image

PR #2650, opened 2025-05-19
Close #2611 trip inventory query

UnSreypich:
In this PR, I have created a query for getting information about the inventory available and total seats for trip_ids that has been passed as the parameter.


channainfo (changes requested):

channainfo:
Add demo of the rspec output.

spec/queries/spree_cm_commissioner/inventory_query_spec.rb
UnSreypich, line 1: For inventory query, I have used the generic name so that it can be used for any variants and can be used with other types of products besides transit.

PR #2673, opened 2025-05-21
Close #2672 sync updated event date to old line items

### LINKED ISSUE #2672: Sync updated event date to old line items

theachoem:
## Problem
When an event changes the date, previous line items still keep the old date, our solution is to manually update the date for all line items.

## Solution
When event dates change, sync them to previous line items.

## In this PR
- [x] Sync new event date to previous line items
- [x] Line item no longer use section dates
  - We use event date directly, and if admin want to override, we can override in variant
  - Hide date fields & other related fields from admin section UI to avoid confusion.

| |
| - |
| Section UI Before |
| image |
| Section UI Now |
| image |

### DIFF
--- a/app/models/spree_cm_commissioner/taxon_decorator.rb
+++ b/app/models/spree_cm_commissioner/taxon_decorator.rb
@@ -46,6 +46,8 @@ module SpreeCmCommissioner
       base.before_save :set_kind
       base.before_save :set_slug
 
+      base.after_update :sync_event_dates_to_line_items, if: -> { saved_change_to_from_date? || saved_change_to_to_date? }
+
       base.whitelisted_ransackable_associations |= %w[vendor]
       base.whitelisted_ransackable_attributes |= %w[kind from_date to_date]
 
@@ -54,6 +56,8 @@ module SpreeCmCommissioner
       base.has_many :invite_user_events, through: :user_events, class_name: 'SpreeCmCommissioner::InviteUserEvent'
 
       base.has_many :line_items, through: :products
+      base.has_many :event_line_items, class_name: 'Spree::LineItem', foreign_key: :event_id
+
       base.has_many :event_blazer_queries, class_name: 'SpreeCmCommissioner::TaxonBlazerQuery'
       base.has_many :blazer_queries, through: :event_blazer_queries, class_name: 'Blazer::Query'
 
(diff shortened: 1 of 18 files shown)


app/models/spree_cm_commissioner/taxon_decorator.rb
```diff
@@ -121,6 +125,10 @@ def send_transaction_email_to_organizers
 
       SpreeCmCommissioner::OrganizersTransactionalEmailNotifier.call(event_id: id)
     end
+
+    def sync_dates_to_line_items
+      ::SpreeCmCommissioner::EventLineItemsDateSyncerJob.perform_later(id)
```
channainfo, line 130: @theachoem Let's verify the condition here to determine if this specific taxon is an event that needs to be synced early. This will help avoid unnecessarily enqueuing a job, as enqueuing a job comes with a cost.

channainfo:
@theachoem Let's create a new ticket to enhance this in the organizer portal as well. Once an event is created, we should prevent editing sensitive fields like the event date. Let's add a "Reschedule Event" feature that allows the organizer to perform the changes you've implemented. There may also be other side effects to address, such as:

* Listing the previous line items.
* Sending an email notification about the rescheduling to the affected line items.
* Resending new tickets based on the previous line items.
* Invalidating old tickets to ensure they are no longer valid.
* Adding a reschedule status (e.g., customer informed and accepted).

Rescheduling happens frequently with bus events as well, and if done without properly informing customers, it could lead to backlash. For example, if a bus is rescheduled to be one hour late and customers aren't notified, they may miss their next planned trip, like connecting flights. It's important to handle this carefully to avoid customer dissatisfaction.

app/models/spree_cm_commissioner/taxon_decorator.rb
theachoem: Fixed bong!

theachoem:
> @theachoem Let's create a new ticket to enhance this in the organizer portal as well. Once an event is created, we should prevent editing sensitive fields like the event date. Let's add a "Reschedule Event" feature that allows the organizer to perform the changes you've implemented. There may also be other side effects to address, such as:
> 
> * Listing the previous line items.
> * Sending an email notification about the rescheduling to the affected line items.
> * Resending new tickets based on the previous line items.
> * Invalidating old tickets to ensure they are no longer valid.
> * Adding a reschedule status (e.g., customer informed and accepted).
> 
> Rescheduling happens frequently with bus events as well, and if done without properly informing customers, it could lead to backlash. For example, if a bus is rescheduled to be one hour late and customers aren't notified, they may miss their next planned trip, like connecting flights. It's important to handle this carefully to avoid customer dissatisfaction.

Ok bong, I will inform to to organizer team.

PR #2730, opened 2025-06-03
Close #2727 make price belong to polymophic

### LINKED ISSUE #2727: Make Spree::Price to belong to polymophic instead of just variant
For reuse in inventory item.

theachoem:


theachoem:
We use this instead:
https://github.com/channainfo/commissioner/pull/2735

